PR #1, opened 2015-01-08
Sw1745 trace check condor ver

brianhlin:


brianhlin:
Since I was tooling around in condor_ce_trace, I noticed a few things with pylint so decided to fix them.

bbockelm:
Looks good to me!

PR #2, opened 2015-01-09
Sw1750 rename job router tool

brianhlin:


bbockelm:
The implementation is a bit distasteful - I hate stabbing around with "which".

However, I don't see any other simple path.  Let's do this for now and visit dropping the "_tool" variant in HTCondor 8.4.

brianhlin:
Well, the 'tool' variant is already dropped in 8.3.x. I thought about doing version comparisons but if we're doing that, we may as well have version requirements in the spec file. The '_tool' variant is already dropped in 8.3.x, the problem is that since we don't have HTCondor version requirements in the CE, we don't know if they're running the latest CE software with an HTCondor version < 8.3.x.

bbockelm:
Yeah - that's why I merged it.  It's really the only viable approach, as much as I dislike "which".

osg-cat:
After reading the comments and looking at the (new) condor_ce_job_router_info code, I do not understand why a fairly simple check of the output of “condor_version” is not the right way to go? Yes, the version string would have to be matched with a grep string, but it’s really not that hard to get right, is it?

bbockelm:
The downsides of version-based:
1) Didn't OSG patch in this tool for a given version - but UW did not?  Hence, condor_version may not be a reliable indicator.
2) We've at least seen one site where two versions of HTCondor were in $PATH; that means parsing "condor_version" would also be unreliable.

osg-cat:
> The downsides of version-based:
> 1) Didn't OSG patch in this tool for a given version - but UW did not? Hence, condor_version may not be a reliable indicator.

Not sure about that – I will have to check with Tim T., probably.

> 2) We've at least seen one site where two versions of HTCondor were in $PATH; that means parsing "condor_version" would also be unreliable.

Well, if we invoke “condor_version” with no path component and then we invoke condor_ce_* also with no path component, wouldn’t both executables come from the same HTCondor directory, even if more than one HTCondor was in $PATH? (Assuming that we picked the right condor_ce_* name based on the version, of course.) I guess I can’t quite see where the unreliability would come from, but I am probably missing something.

— Tim=

brianhlin:
Tim and I had a discussion this morning:

* If a user has two different versions of HTCondor installed with 8.3.x later in the path, the current version of the CE tool will use condor_job_router_info. This behavior is slightly different than our other wrapper scripts that use the binaries that are first in path. Instead, we should `which` for something that should be in every install (i.e., condor_version) and construct the path to the executables that way.
* We should probably use /bin/which in case the user has any aliases for which

edquist:
One gotcha -- if you are going to use `which` you should use /usr/bin/which, not /bin/which.  /usr/bin is the standard location, and, at least on RHEL6, there is no /bin/which symlink. (I get a `bash: /bin/which: No such file or directory` error.)

PR #5, opened 2015-02-07
Updated functionality for Syracuse OSG site.

bbockelm:
This provides two new pieces of functionality:
1) Allow the site-level HTCondor-CE collector to accept startd ads
   from worker nodes; advertise the collector's address in the job
   environment.  This is meant to work in conjunction with a new
   gWMS feature in 3.2.9 / 3.2.10 (exact release unknown) which advertises
   the pilot startd ad.  This will enable Miron's long-standing
   request to make pilot ads visible to the CEs.
2) Implement a new attribute, minWalltime, which specifies a minimum
   valid runtime for a job; if the pilot runs for less than this amount
   of time, it is considered to be a problematic job and put on hold.
   This will allow site admins more visibility into problem pilots and
   worker nodes.

### DIFF
--- a/config/condor_mapfile
+++ b/config/condor_mapfile
@@ -1,6 +1,6 @@
 GSI "^\/DC\=com\/DC\=DigiCert-Grid\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
 GSI "^\/DC\=DigiCert-Grid\/DC\=com\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
 GSI (.*) GSS_ASSIST_GRIDMAP
-GSI (.*) anonymous@gsi
+GSI (/CN=[-.A-Za-z0-9/= ]+) \1@unmapped.opensciencegrid.org
 CLAIMTOBE .* anonymous@claimtobe
 FS (.*) \1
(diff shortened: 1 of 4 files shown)


brianhlin:
Looks good.

config/condor_mapfile
```diff
@@ -1,6 +1,6 @@
 GSI "^\/DC\=com\/DC\=DigiCert-Grid\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
 GSI "^\/DC\=DigiCert-Grid\/DC\=com\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
 GSI (.*) GSS_ASSIST_GRIDMAP
-GSI (.*) anonymous@gsi
+GSI (/CN=[-.A-Za-z0-9/= ]+) \1@unmapped.opensciencegrid.org
```
brianhlin: @bbockelm, I'm noticing errors in my SchedLog with the newest version of HTCondor CE:

> 03/17/15 16:29:10 ERROR: Error compiling expression '(/CN=[-.A-Za-z0-9/=' -- missing terminating ] for character class.  this entry will be ignored.

What is this regex supposed to be matching? should we replace the ' ' with '\s' to stop condor from complaining?

bbockelm: Hi Brian,

Actually, I think it's complaining about the "=", isn't it?

This is just supposed to match any reasonable pilot CN.

Brian

brianhlin: Nope, looks like the space acting as the field separator in the mapfile: #15 .

PR #11, opened 2015-02-20
Separate out functions/exceptions used by both condor_ce_run and condor_...

brianhlin:
...ce_trace into their own library. This is just refactoring so it doesn't need to make it into the next release.

### DIFF
--- a/src/condor_ce_tools.py
+++ b/src/condor_ce_tools.py
@@ -5,7 +5,7 @@ import os
 import tempfile
 import textwrap
 import time
-from subprocess import Popen, PIPE, STDOUT
+from subprocess import Popen, PIPE
 
 # Excluding submit file so the respective scripts
 # can generate it as they see fit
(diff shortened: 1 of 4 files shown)


src/condor_ce_tools.py
```diff
@@ -0,0 +1,49 @@
...
+        print line
+    print "*"*80
+
+def print_timestamped_msg(msg):
+    print_formatted_msg("%s %s" % (time.strftime('%F %T'), msg))
+
+def run_command(command):
+    p = Popen(command, stdout=PIPE, stderr=STDOUT)
+    stdout, stderr = p.communicate()
+    return p.returncode, stdout, stderr
```
matyasselmeci: stderr is always going to be blank here because you are redirecting stderr to stdout above. If you want to separate the two streams, you should do
    p = Popen(command, stdout=PIPE, stderr=PIPE)
If not, then don't capture and return stderr, it's just confusing
    outerr, _ = p.communicate()
    return p.returncode, outerr

brianhlin: Ah, that makes sense. We should probably fix it in osg-test as well.

src/condor_ce_trace
```diff
@@ -123,45 +105,45 @@ def parseOpts():
...
-            runPing("<%s:%d>" % addrinfo[4][:2], daemon="COLLECTOR")
-            hadSuccess = True
+            run_ping("<%s:%d>" % addrinfo[4][:2], daemon="COLLECTOR")
+            had_success = True
             print "- Successful ping of collector on %s.\n" % addr
         except Exception, e:
             print "- Failed ping of collector on %s.\n" % addr
             raise e
 
     if not hadSuccess:
```
matyasselmeci: ^ should be "if not had_success:"

PR #27, opened 2015-04-27
Fix bug that prevented HTCondor CE service from starting with multiple job routes

brianhlin:


src/condor-ce
```diff
@@ -50,13 +50,11 @@ start() {
...
 
     # Check validity of job router knobs
-    for ATTR in 'JOB_ROUTER_DEFAULTS' 'JOB_ROUTER_ENTRIES'; do
-        python -c "import htcondor; import classad; classad.ExprTree(htcondor.param[\"$ATTR\"])" &> /dev/null
-        if [ $? -ne 0 ]; then
-            echo "Could not read $ATTR in the HTCondor CE configuration. Please verify syntax correctness."
-            exit 6
-        fi
-    done
+    python -c "import htcondor; import classad; classad.ExprTree(htcondor.param['JOB_ROUTER_DEFAULTS'])" &> /dev/null
```
bbockelm: You should be able to instead use the classad.parseAds method to parse multiple ads.  *note* that I don't quite recall when parseAds was added (but I think it was 8.1.x).

brianhlin: Looks like it first made it into 8.1.2 so that should work!

PR #28, opened 2015-04-28
Re-apply 5f2ae1d with classad.parseAds() instead of

brianhlin:
classad.ExprTree() to support JOB_ROUTE_ENTRIES with multiple
ClassAds. Non-existant attributes or syntatically incorrect ClassAds
will throw exceptions


brianhlin:
This solution is incomplete since it only checks the first entry in the JOB_ROUTER_ENTRIES. There doesn't seem to be an easy way to verify the JOB_ROUTER_ENTRIES so I'm closing this.

bbockelm:
Hi Brian,

Why not include the incomplete solution and do the complete one later?

I'm working on a revamp of the parseAds (and friends) for the 8.5.x series, and that probably will do precisely what you want.

Brian

brianhlin:
I was hoping we could find a full solution instead of applying a half-baked one. I suppose it doesn't hurt anything to have it in though so I can submit a new PR if you'd like.

bbockelm:
A half-baked cake probably tastes better than no cake at all!  Indeed, a fully baked one will be better...

I just don't see how to make this bulletproof without a little more help from the htcondor side.

Yeah, let's reopen.

brianhlin:
Oops, didn't mean to reclose this

brianhlin:
@matyasselmeci Could you take a look when you have a chance? I did a quick test on a fermicloud VM and all seemed to work well to me.

PR #29, opened 2015-06-08
Add COLLECTOR_REQUIREMENTS expression to ce-collector config

matyasselmeci:
This expression keeps rogue sites from being able to send CE ads, or
xrootd startd/master ads for other sites.  This should get rid of the
need for a whitelist of sites that can send ads.

This is for SOFTWARE-1790 / SOFTWARE-1820.


bbockelm:
Looks good to me.

PR #32, opened 2015-06-19
Add 04-ce-collector-auth.conf for the ce-collector install (SOFTWARE-1790)

matyasselmeci:
This allows all sites to send schedd, startd, and master ads to the CE
collector, instead of just those in a generated whitelist.  This is safe
because the COLLECTOR_REQUIREMENTS expression introduced in 6679194 keeps sites
from overwriting other sites' ads.


bbockelm:
Looks good to me.

PR #36, opened 2015-08-07
Add default_remote_cerequirements

djw8605:
Add default_remote_cerequirements.  This enables the administrator to
specify routes with the attribute default_remote_cerequirements, which
will then be appended to the remote_cerequirements.  Therefore, to add
a new type of resource (such as a gpu), the admin needs to:

1. Add to the default_remote_cerequirements in the router entry so
that the local submit file (.sh) can pick up the new requirement.  For
a GPU, it would be:

```
set_default_remote_cerequirements = "RequestGpus == 1"; \
```

2. Add to the [pbs|lsf|sge|…]_local_submit_attributes.sh an if
statement to capture and use this attribute.  Again, for a GPU and
Slurm:

```
if [ -n "$RequestGpus" ]; then
    echo "#SBATCH --gres=gpu:$RequestGpus"
    echo "module load cuda/6.0"
fi
```


djw8605:
@brianhlin, could you take a look at this pull request.

brianhlin:
Looks good to me

PR #37, opened 2015-08-18
htcondor-ce-collector should source /etc/sysconfig/condor-ce-collector

brianhlin:
instead of trying to source /etc/sysconfig/condor-ce


brianhlin:
@matyasselmeci Could you take a look when you have a chance?

PR #40, opened 2015-09-01
Try to determine the address HTCondor would use.

bbockelm:
This tries to replicate the logic of the "default address" HTCondor will use.

We want to verify the forward/backward mapping of this address matches the hostname on the certificate.  Otherwise, the child daemon may not authenticate the master.

### DIFF
--- a/src/condor_ce_host_network_check
+++ b/src/condor_ce_host_network_check
@@ -136,9 +136,9 @@ def iface_matches(network_iface, pattern):
     return False
 
 def is_local_addr(ipv4_addr):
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Nitpicks aside, it looks good. Has it been tested?

bbockelm:
Yeah - I tried various misconfigurations and I think I hit all the different code paths.

osg-cat:
I would still like Mat (or someone) to do a code review / hardening session with it, before it becomes a standard tool.

— Tim

PR #55, opened 2015-11-16
Adding a web monitor

djw8605:
This pull request adds a web monitor.  The web monitor is run as a HTCondor daemon serving requests.  It generates RRDs and summary json information stored on disk which is then served on demand.

A single CE view.  This is default:
image

Multi-CE view.  This can be configured:
image


djw8605:
@bbockelm If you are happy with the current version, lets merge this before the December development freeze.

bbockelm:
Can you confirm that the monitor is turned off by default?  If so, I'm happy.

djw8605:
Yes, in the [ce-webapp.conf](https://github.com/opensciencegrid/htcondor-ce/pull/55/files#diff-a6667b407ea70b9f709318685cb027f8R12) file, the daemon list is commented out by default.

But... the cron is active by default.  Is that desired?

bbockelm:
The cron should no-op itself if the webapp isn't in the daemon list.  Should be easy to detect?

djw8605:
I've added a no-op in the cron if CEMON is not in DAEMON_LIST (see 1171f3c)

bbockelm:
Sounds good.  Anything else in particular I should pay attention to before merging?

djw8605:
I can't think of much else.  1 question though, [in jobs display](https://github.com/djw8605/htcondor-ce/commit/20199e846baab961dd57ce401e11428abe181502#diff-4188525908cff7d8df84341c8265e43dR333), is that proper way to parse arguments?  I think so.

djw8605:
@bbockelm I've moved the import to the top of the file in 1d485e4

brianhlin:
@djw8605: [This if statement](https://github.com/opensciencegrid/htcondor-ce/blob/master/config/05-ce-webapp-defaults.conf#L45-L48) could be problematic for versions of HTCondor < 8.2. I'm not sure if this is a big deal but if we want to address this, you can throw `:` in front of each of those lines like in [03-ce-shared-port.conf](https://github.com/opensciencegrid/htcondor-ce/blob/master/config/03-ce-shared-port.conf#L18-L23).

I also assume that even though the CEMON (Are we actually going with that name? We really should change it to avoid confusion with the old CEMon) is in the `DC_DAEMON_LIST` it won't be started since it's not in the `DAEMON_LIST` by default, correct?

djw8605:
@bbockelm I'm out the rest of this week, can you take a look at BrianL's comments?

PR #56, opened 2015-11-25
Changing name from webapp and CEMON to CEVIEW.

djw8605:


djw8605:
@bbockelm I updated the name from CEMON to CEVIEW.  Could you do a quick review.  I have a test instance of an RPM with these changes running at http://red-foreman.unl.edu:4440/

bbockelm:
Looks good to me.

PR #60, opened 2015-12-09
Add BDII providers for HTCondorCE, for non-OSG sites

IainSteers:
Here are a couple of info providers for BDII GLUE1.3 and GLUE2 publication.

Handles multi-CE sites and allows users to specify their site-specific values inside HTCondor-CE config, meaning that there's no need for extra config outside of the CE.

I've left the spec changes out as the one in head seems out-of-date. I'm envisioning a bdii sub-package, which has a Require on bdii.

ifdef'd out for OSG builds.

Providers would need to be installed in /var/lib/bdii/gip/provider/ owned by the ldap user/group.

Let me know if you need any changes or have any comments/corrections.


config/06-ce-bdii.conf
```diff
@@ -0,0 +1,20 @@
...
+
+# For Zookeeper
+#HTCONDORCE_BDII_ELECTION = ZOOKEEPER
+#HTCONDORCE_BDII_ZKHOSTS =
+
+# BDII Static Info and VOs
+HTCONDORCE_VONames = atlas, cms, alice, lhcb, dteam
+HTCONDORCE_SiteName = SITENAME
+HTCONDORCE_HEPSPEC_INFO = 10.00-HEP-SPEC06
+HTCONDORCE_CORES = 16
```
kreczko: Hi Iain. Just a quick question: `HTCONDORCE_CORES` refers to the total amount of cores in the cluster, is that correct?

IainSteers:
Hi @kreczko, no HTCONDORCE_CORES refers to the number of Logical CPUS in a single execute node.

The total cores in the cluster is calculated by querying the COLLECTOR_HOST using the python bindings. 

The reason it's hard-coded here is that some people run vastly differing hardware setups in their pools, e.g. most nodes might be 8 core, but they may have a large number of 32-core nodes.

We could average by querying node cpus through the collector but then we'd need to floor it and then it doesn't exactly reflect reality. Rather than that approach I decided to leave it up to the site admin.

Happy to come up with a better approach if you have ideas.

brianhlin:
Iain, I only merged in some of the spec file changes from the OSG 3.3 branch because other parts of it don't apply to 3.2. If you include your version of the spec file, I can help you work on it so that it can be merged in cleanly.

IainSteers:
Hi @brianhlin, no problem. I'll submit the spec changes in a couple of hours.

IainSteers:
Hi @brianhlin, this is a preliminary spec change for a bdii sub-package. 

Let me know if something doesn't quite look right. I wasn't sure what kind of tag you'd prefer to ifdef on so that's missing at the moment.

Thanks, Iain

brianhlin:
You can wrap the bdii blocks in:

    %if ! 0{%osg}
        ...
    %endif

IainSteers:
Hi @brianhlin does this look reasonable?

edquist:
@brianhlin,

I just noticed a gotcha in your last comment; you should use:

    0%{?osg}

instead of:

    0{%osg}

Otherwise it will not expand the way you'd like...

IainSteers:
@edquist thanks for the spot!

I completely missed that.

brianhlin:
Thanks @edquist. Excuse my brain fart, please D:

brianhlin:
@IainSteers: Spec file looks good, would you mind squashing all the spec commits? I'll merge after you do that and add a comment for `HTCONDORCE_CORES`

IainSteers:
@brianhlin Sure!

Give me two ticks and I'll squash them in.

IainSteers:
@brianhlin Done!

config/06-ce-bdii.conf
brianhlin: Could you change the last line to something like the following?

    HTCONDORCE_CORES = 16 # Cores per node

brianhlin:
Looks good.

brianhlin:
@IainSteers Does the `bdii` package provide the `/var/lib/bdii/gip/provider` dir?

IainSteers:
@brianhlin It does yes, do you think it would be safer if the rpm install did an mkdir anyway?

brianhlin:
It should be fine, I just wanted to make sure that the HTCondor-CE package didn't have to own any of those dirs.

IainSteers:
Hi @brianhlin just looked back at the spec and realised I put the ownership of the two provider scripts as root from the looks of it.

It actually needs to be the ldap user and group which is created by the bdii package.

ls -lah /var/lib/bdii/gip/provider/condor_ce_bdii_generate*
-rwxr-xr-x. 1 ldap ldap  14K Sep 18 08:28 /var/lib/bdii/gip/provider/condor_ce_bdii_generate_glue2.py
-rwxr-xr-x. 1 ldap ldap 9.0K Sep 18 08:28 /var/lib/bdii/gip/provider/condor_ce_bdii_generate_glue1.py

brianhlin:
I'm working on some spec file changes atm so I'll fix that while I'm at it.

PR #62, opened 2015-12-14
CEVIEW review

brianhlin:
@djw8605 This is a WIP but figured I'd get this up sooner rather than later so you can take a look at your convenience. The main question I've got, which perhaps @bbockelm may have a better answer for, is about the use of `__import__` in jobmetrics. I didn't really see a reason for it but if there was one, I could easily revert that commit.


bbockelm:
The __import__ is done within the functions because we want to avoid loading HTCondor code in the parent and forking.

I had some hard-to-track issues that originally appeared to be caused by the condor code having globals that didn't re-initialize correctly in the child process.

brianhlin:
We should drop this then, https://github.com/opensciencegrid/htcondor-ce/blob/master/src/condor_ce_jobmetrics#L15, since the only function that uses it is satisfied with an `__import__`. Sound good to you?

brianhlin:
Also, @edquist brought up a good point that `import` within the functions should maintain scope just as `__import__` does. Any reason in choosing one over the other here?

bbockelm:
Just a matter of habit.  As long as it doesn't get done in the parent, either syntax works for me.

PR #68, opened 2016-02-17
Testing failures

brianhlin:
Ensuring @djw8605's tests fail on a malformed spec file.


djw8605:
Yay, failures are actually failures!

brianhlin:
Great!

PR #69, opened 2016-02-22
Drop all RSL references (SOFTWARE-2205)

brianhlin:
A few spots with references to the RSL/classad libraries were missed. Build passed here: https://travis-ci.org/brianhlin/htcondor-ce/builds/111019051

v2.0.1 should be ok though even if we leave the JOB_ROUTER_DEFAULTS RSL functions in there because of all the null value handling.

### DIFF
--- a/config/htcondor-ce.spec
+++ b/config/htcondor-ce.spec
@@ -136,8 +136,6 @@ Group: Applications/System
 Summary: Client-side tools for submission to HTCondor-CE
 
 BuildRequires: boost-devel
-BuildRequires: globus-rsl-devel
-BuildRequires: condor-classads-devel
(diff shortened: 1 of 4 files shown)


djw8605:
It looks like the tests are still installing globus components? 
https://travis-ci.org/opensciencegrid/htcondor-ce/jobs/111026248#L1339

brianhlin:
That's not coming from HTCondor-CE, that's actually coming from HTCondor's dep resolution.

https://travis-ci.org/opensciencegrid/htcondor-ce/jobs/111026248#L1133-L1153

djw8605:
Ok, then looks good.  Merging.

PR #71, opened 2016-03-22
Accept subject DNs in extattr_table.txt (SOFTWARE-2243)

brianhlin:
This patch allows HTCondor-CE to use lines out of `/etc/osg/extattr_table.txt` to be matched against the user's subject DN in addition to their VOMS attributes. Instead of [x509UserProxyFirstFQAN](http://research.cs.wisc.edu/htcondor/manual/v8.4/12_Appendix_A.html#100112), we use [x509UserProxyFQAN](http://research.cs.wisc.edu/htcondor/manual/v8.4/12_Appendix_A.html#100116), which includes the subject DN as well as the VOMS FQAN in a single attribute.

`condor_ce_trace` was also fixed to generate an `x509UserProxyFQAN` that mimics the format of the attribute created via `condor_submit`.

### DIFF
--- a/src/condor_ce_router_defaults
+++ b/src/condor_ce_router_defaults
@@ -139,10 +139,15 @@ def set_accounting_group():
         return None
     accounting_group_str = ''
     for mapping in uid_mappings:
-        accounting_group_str += 'ifThenElse(Owner == %s, strcat(%s, ".", Owner), ' % (quote(mapping[0]), quote(mapping[1]))
+        accounting_group_str += 'ifThenElse(Owner == %s, strcat(%s, ".", Owner), ' % (quote(mapping[0]),
+                                                                                      quote(mapping[1]))
     for mapping in attr_mappings:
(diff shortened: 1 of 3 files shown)


djw8605:
Does this change the formatting of the extattr_table.txt if users have already created one?

bbockelm:
What happens if the table matches a secondary group?

I think we should first look at the DN, then the primary FQAN, but not all FQANs.

Brian

brianhlin:
@bbockelm: I've added a commit that should address your concerns (and also fix the case when VOMS attributes weren't defined).

@djw8605: This should not change the format for pre-existing `extattr_table.txt` files.

bbockelm:
Dumb question, but can we add a unit test for this?

brianhlin:
Do you want something added to osg-test or do you just want unit tests in the CE source?

brianhlin:
I spoke with @bbockelm about the unit tests and he said that writing one isn't a blocker. Anyone have any other concerns or can I merge this?

djw8605:
Yeah, no need for a unit test right now, lets just pull it.

PR #75, opened 2016-04-05
Adding Schedd to the CEView for the Schedd Cron

djw8605:
Adding the Schedd for dedicated ce-collectors and ce-view machines.
The Schedd is necessary for the Schedd cron which runs the metrics
updates.  Additionally, add example configuration which will turn
off all job submission abilities to this schedd.

SOFTWARE-2268

### DIFF
--- a/config/05-ce-view.conf
+++ b/config/05-ce-view.conf
@@ -11,8 +11,8 @@
 # To enable the webapp, uncomment the line below:
 # DAEMON_LIST = $(DAEMON_LIST), CEVIEW, GANGLIAD, SCHEDD
 
-# If this is only a CE View, or a ce-collector, then uncomment below to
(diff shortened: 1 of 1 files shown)


brianhlin:
What exactly is the 'schedd cron'? Does it run as a job?

djw8605:
The Schedd Cron does not run as a job, but instead as a process that runs at certain intervals to update the metrics.  Here is the configuration: https://github.com/opensciencegrid/htcondor-ce/blob/master/config/05-ce-view-defaults.conf#L33

brianhlin:
Looks good to me then. Have you tested the change?

djw8605:
Yup.
screenshot-10 71 101 206 2016-04-05 10-59-06

PR #77, opened 2016-04-05
Accept submit attribute format, '+AttributeName'

brianhlin:
We tell users in the [submitting jobs document](https://twiki.opensciencegrid.org/bin/view/Documentation/Release3/SubmittingHTCondorCE) that they need to prepend some of their attributes with '+'. Before this change, `condor_ce_trace` would fail on such attributes.


bbockelm:
@brianhlin - any reason why this hasn't been merged already?

brianhlin:
Nope, just needs a review.

brianhlin:
@djw8605: could you check this out when you get a chance?

PR #81, opened 2016-04-14
Adjust hold message for reaching MaxJobs or MaxIdleJobs

brianhlin:
We've been seeing this more among sites. I don't think there's a way to differentiate between the job not matching any routes or the routes being full from the JobAd so this is better than nothing.


djw8605:
Seems reasonable to me.

PR #82, opened 2016-04-18
Add child as authorized daemon.

djw8605:
Fixes [SOFTWARE-2280](https://jira.opensciencegrid.org/browse/SOFTWARE-2280):

> The GridManager needs permissions to file transfers. Current fix is to add condor@child/$(FULL_HOSTNAME) to FRIENDLY_DAEMONS from 01-ce-auth.conf.

### DIFF
--- a/config/01-ce-auth-defaults.conf
+++ b/config/01-ce-auth-defaults.conf
@@ -15,7 +15,7 @@ DENY_ADMINISTRATOR = anonymous@*, unmapped@*
 DENY_DAEMON = anonymous@*, unmapped@*
 
 # Defaults authorizations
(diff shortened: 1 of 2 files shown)


brianhlin:
What testing has this undergone?

djw8605:
I've run it in htcondor-ce-bosco.  I know it works in that case, but I'm not sure of the security implications.

brianhlin:
Got a thumbs up from Zach Miller on this and an explanation of `condor@child`:

> condor@child is a security session set up automatically between child and parent processes.  There's also a condor@parent principal.
>
> Examples: commands from the Shadow to the Schedd would be authorized as condor@child in the Schedd.  Commands from the Schedd to the Master (like the "I'm still alive" message) would also be condor@child in the Master.  A command from the Master to the Schedd (like "Shutdown now") would be authorized in the Schedd as "condor@parent".

PR #85, opened 2016-04-19
Adding yum clean all to clear the yum cache before test

djw8605:
To fix the failing builds.  Oh at the failing builds!


brianhlin:
Looks ok to me

PR #86, opened 2016-04-20
Add expire-cache cleaning since it's skipped with 'yum clean all'

brianhlin:
Even though the 3.2.29 yum man page says `yum clean all` performs all of the clean options, closer inspection of the client shows that it skips the `expire-cache` option. We perform the same two  commands in [osg-test](https://github.com/brianhlin/osg-test/blob/master/osgtest/library/yum.py#L16).

### DIFF
--- a/tests/test_inside_docker.sh
+++ b/tests/test_inside_docker.sh
@@ -6,6 +6,7 @@ ls -l /home
 
 # Clean the yum cache
 yum -y clean all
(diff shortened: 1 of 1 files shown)


brianhlin:
Though, the original problem of bad connectivity to yum repos isn't solved by either of the `yum clean` commands.

djw8605:
Never a bad idea.

PR #88, opened 2016-04-27
Enable osg-test run for travis-ci PRs.

bbockelm:
Let's see if osg-test can run quickly enough to make sense for the CI build.


bbockelm:
Alright, we're green across the board.  Time for a squash-commit.

PR #91, opened 2016-04-27
Split out test setup so we can start system services

djw8605:
Split out the test setup so we can start system services using systemd or regular init.

### DIFF
--- a/tests/setup_tests.sh
+++ b/tests/setup_tests.sh
@@ -17,9 +17,9 @@ DOCKER_CONTAINER_ID=$(docker ps | grep centos | awk '{print $1}')
 docker logs $DOCKER_CONTAINER_ID
 docker exec -ti $DOCKER_CONTAINER_ID /bin/bash -c "bash -xe /htcondor-ce/tests/test_inside_docker.sh ${OS_VERSION};
   echo -ne \"------\nEND HTCONDOR-CE TESTS\n------\nSystemD Units:\n------\n\"; 
-  systemctl --no-pager --all --full status
+  systemctl --no-pager --all --full status;
   echo -ne \"------\nJournalD Logs:\n------\n\" ;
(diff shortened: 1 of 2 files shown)


djw8605:
@brianhlin Could you review and merge this.  It is needed by #88 to perform tests on el6 and 7.

tests/setup_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ "$el_version" = "6" ]; then
+
+sudo docker run --rm=true -v `pwd`:/htcondor-ce:rw centos:centos${OS_VERSION} /bin/bash -c "bash -xe /htcondor-ce/tests/test_inside_docker.sh ${OS_VERSION}"
+
+elif [ "$el_version" = "7" ]; then
+
+docker run --privileged -d -ti -e "container=docker"  -v /sys/fs/cgroup:/sys/fs/cgroup -v `pwd`:/htcondor-ce:rw  centos:centos${OS_VERSION}   /usr/sbin/init
+DOCKER_CONTAINER_ID=$(docker ps | grep centos | awk '{print $1}')
+docker logs $DOCKER_CONTAINER_ID
+docker exec -ti $DOCKER_CONTAINER_ID /bin/bash -c "bash -xe /htcondor-ce/tests/test_inside_docker.sh ${OS_VERSION};
```
bbockelm: What happens if this script fails?

djw8605: At the top, the argument is `#!/bin/bash -xe`, the `e` of which means it will it will exit at the first non-zero exit command, with that status.  So, if this script fails, the build will fill.

brianhlin:
Looked over the Travis CI test output and it looks like everything ran as expected.

tests/setup_tests.sh
bbockelm: Ok, sounds good.

PR #93, opened 2016-04-28
Adding -xe to bash command for testing

djw8605:
Adding the `-xe` command for testing so that if any of the command exits with non-zero status, all of it will exit.


djw8605:
The systemctl command seems to output non-zero exit status.  I removed it, as well as the jounal command.  These are debugging commands that we copied from the ansible project.

@bbockelm could you review and merge this.

bbockelm:
Looks good to me.

PR #96, opened 2016-05-17
condor_ce tools should print exception to stdout when using -verbose …

brianhlin:
…(SOFTWARE-2211)

- I had to adjust the printing slightly so that the traceback formatting doesn't break.
- Some variable renames to appease pylint

### DIFF
--- a/src/condor_ce_trace
+++ b/src/condor_ce_trace
@@ -348,19 +348,24 @@ def main():
 if __name__ == '__main__':
     try:
         main()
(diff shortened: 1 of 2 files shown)


brianhlin:
@djw8605: I've tested this and see the expected behavior:

1. Without the `--debug` flag, the exception is output to a file but hidden from stdout.
1. With `--debug` specified, the exception is printed to stdout as well as to the `.stack*` file.

Could you review this change?

PR #98, opened 2016-06-02
Execute osg-test in subshell so we can get logs

djw8605:


brianhlin:
Won't this still exit immediately if osg-test fails? Perhaps we should drop the `-x` argument to the shebang line?

djw8605:
We would need to drop the `-e` argument.  But you are right, subshell doesn't do it.

PR #99, opened 2016-06-02
Change 'set' args around osg-test so we can get logs

brianhlin:


djw8605:
I kinda wish the build would have failed to fully the change.  This change looks good to me.

brianhlin:
If you want, I can throw in a failing command in between the `set` commands to see it fail on a separate branch.

djw8605:
Sure!

brianhlin:
Added a failing command in between the `set` commands and we still got the logs back! 

https://travis-ci.org/brianhlin/htcondor-ce/builds/135056411

PR #102, opened 2016-08-02
Force 'condor_ce_q -allusers' until QUEUE_SUPER_USER is fixed to be able

brianhlin:
to use CERTIFICATE_MAPFILE in 8.5.6 (SOFTWARE-2412)


brianhlin:
Tested this with different versions of 8.5 and it works as expected.

PR #106, opened 2016-08-24
Properly handle unmapped GSI users (SOFTWARE-2433)

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2433

### DIFF
--- a/config/01-ce-auth-defaults.conf
+++ b/config/01-ce-auth-defaults.conf
@@ -1,4 +1,3 @@
-
 ###############################################################################
 #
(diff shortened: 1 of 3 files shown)


brianhlin:
@bbockelm: I don't think I'm missing anything here but I just want to double check your intent of allowing unmapped users to advertise to the collector.

brianhlin:
To avoid messing with pilots advertising to the CE, we'll leave this alone. Instead, we will solve the original issue by adding the missing CA line to {{condor_mapfile}}.

PR #108, opened 2016-08-24
Fix OnExitHold to be set to expressions rather than their evaluated f…

brianhlin:
orms. TJ noticed this when putting together the classad transform language. @bbockelm, could you review this since you did the original work on `OnExitHold`?


brianhlin:
@bbockelm gave this the thumbs up.

PR #109, opened 2016-08-29
Remove the HTCondor-CE init script on EL7 (SOFTWARE-2419)

brianhlin:
- EL7 tmpfs support
- Run against `osg-test` trunk for tests that accommodate this PR


brianhlin:
@axs-gentoo: Hey Ian, since you submitted #103 for tmpfs dir creation, would you be able to check out `config/condor-ce.conf` and `config/condor-ce-collector.conf` in 5b47d07 and verify that they look ok? I haven't worked with systemd-tmpfiles before.

djw8605:
Read through the changes, looks good to me.

axs-gentoo:
I concurr -- these look good.  I didn't see a need for /var/lock/condor-ce/user when I was running into issues, but it won't hurt to have it there for sure.  Note also that we should likely file a condor-cron.conf for that project too, to match.

brianhlin:
Ticket for condor-cron created [here](https://jira.opensciencegrid.org/browse/SOFTWARE-2439).

PR #111, opened 2016-08-31
Fix rpm cleanup on EL7

brianhlin:


brianhlin:
Added commits with tests in #112

PR #117, opened 2016-09-22
Handle unbounded HTCondor-CE accounting dir (SOFTWARE-2090)

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2090

I've made the schedd cron only apply to cases of non-HTCondor backends (checking for `JOB_ROUTER_SCHEDD2_NAME`). I've also disabled `PER_JOB_HISTORY` for HTCondor backends per [this comment](https://jira.opensciencegrid.org/browse/SOFTWARE-2463?focusedCommentId=339415&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-339415).

The build fails because it requires a bug fix available in 8.4.9 and that's not available in any of the enabled repos. [Here's a set of tests](https://travis-ci.org/brianhlin/htcondor-ce/builds/162047276) that passed when that specific version of HTCondor was available.

@bbockelm, feel free to review this as well.


brianhlin:
@djw8605 : Could you review this Monday?

src/gratia_cleanup.py
```diff
@@ -0,0 +1,17 @@
...
+
+import os
+import datetime
+
+os.environ['CONDOR_CONFIG'] = '/etc/condor-ce/condor_config'
+
+import htcondor
+
+GRATIA_DIR = htcondor.param['PER_JOB_HISTORY_DIR']
+HISTORY_FILES = [os.path.join(GRATIA_DIR, x) for x in os.listdir(GRATIA_DIR)]
```
djw8605: I'm a bit worried about listing an entire directory, if it gets out of hand.  What if there are lots, and lots of files in this directory.

djw8605:
I see the checks are failing because of requiring a newer version of HTCondor.  When will 8.4.9 be in the OSG repos?

brianhlin:
The pre-release build is in osg-development and we'll get the released version this Thursday.

djw8605:
Should we include osg-development in the build/test matrix for travis-ci?  Or osg-testing?

brianhlin:
I could see that leading to a lot of failed tests in the future I'm not sure we gain much by enabling development. Testing is reasonable, though, even if it won't solve our failing tests until Thursday.

src/gratia_cleanup.py
brianhlin: How much is 'lots and lots'? I just touched 100k files in the history dir, ran the script (modified to remove files older than a second), and the script completed in 2.4s.

djw8605:
eh, it's probably ok.

PR #121, opened 2016-10-11
Handle '-local-name' option in the CE View

brianhlin:
Background: https://jira.opensciencegrid.org/browse/SOFTWARE-2477

Also some spec file updates. I tested the option handling and verified that the CE View page now works upon service restart.

### DIFF
--- a/src/condor_ce_view
+++ b/src/condor_ce_view
@@ -2,7 +2,7 @@
 
 import os
 import pwd
-import errno
+import sys
 import socket
 import logging
 import optparse
@@ -81,7 +81,7 @@ def setup_logging():
     cherrypy.log.access_log.addHandler(h)
 
 
-def parse_opts():
+def parse_opts(**kwargs):
     parser = optparse.OptionParser()
     parser.add_option("-p", "--pool", help="HTCondor-CE pool to consider.", dest="pool")
     parser.add_option("-n", "--name", help="HTCondor-CE schedd to consider.", dest="name")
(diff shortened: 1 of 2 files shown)


brianhlin:
@edquist the Travis-CI checks will fail because of the version dependencies in the latest CE :(

src/condor_ce_view
```diff
@@ -92,7 +92,10 @@ def parse_opts():
     opts = parser.parse_args()[0]
 
     if hasattr(htcondor, 'set_subsystem'):
-        htcondor.set_subsystem("CEVIEW")
+        try:
+            htcondor.set_subsystem(kwargs['localname'])
+        except KeyError:
+            htcondor.set_subsystem("CEVIEW")
```
edquist: It _seems_ like the intention here is that `localname` is an optional argument.  In that case, the above works, but (to me anyway) it's not clear whether the try/except for a `KeyError` is for `kwargs` or the `set_subsystem` call, or both.

If it's just for `kwargs`, you could test it with `if 'localname' in kwargs` instead of the try/except, or just do a `dict.get` with a default:
```htcondor.set_subsystem(kwargs.get('localname', 'CEVIEW'))```

But if `localname` is the only optional argument for `parse_opts`, you could also just define the function that way with `def parse_opts(localname=None)` and then later test if `localname is None` or be slightly lazy (but not dangerously) and do something like:
```htcondor.set_subsystem(localname or "CEVIEW")```

matyasselmeci:
LGTM.

PR #124, opened 2016-10-18
Accept all DaemonCore options (SOFTWARE-2481). Change '-p' to refer to

brianhlin:
the port to bind to instead of the pool to track to respect DaemonCore.


djw8605:
Looks good to me!

PR #126, opened 2016-11-22
Fix hold msg for jobs not picked up by the JR (SOFTWARE-2539)

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2539

There are at least 3 reasons that the Job Router will ignore idle jobs
in the Schedd for 30 min+:

1. Job did not match any available job routes
2. All potential job routes are full (i.e. reached MaxJobs)
3. Routed jobs are failing and throttling occurs; job submission rates
drop and the job router does not pick up new jobs until the failure
rate drops below the configured threshold

Rather than including them in a very long hold message, it makes more
sense to elaborate the reasons in our documentation.


djw8605:
Just to elaborate, so the hold message "knows" why the job was put on hold.  But we have chosen to not list it?

brianhlin:
Well, we know why the job was held, it wasn't picked up by the job router for 30+ min. The reasons why the JR may have done that are varied and if we listed them all would result in a very long hold message. Plus, if we find another reason that the JR didn't pick up the job in time, we can just update the documentation instead of having to cut a new version of HTCondor-CE.

djw8605:
So our choices where:
1. Give detailed error message, that lists the known possible issues, but may not list all the possible problems.
1. Give an error message that lists none of the problems, and instead says "see troubleshooting docs", and not even a link?

brianhlin:
1. Results in an obscenely long hold message
1. We can add a link but I think it's unnecessary with the correct document being the top link for the 'troubleshooting htcondor-ce' google search

djw8605:
I like error messages that are descriptive.  A long hold (or error) message does not bother me at all, if it describes the problem.  You could even include the text "see troubleshooting docs" in the hold message if necessary.

brianhlin:
@djw8605: I've updated the commit, how's it look now?

PR #127, opened 2016-11-23
Remove TotalSubmitProcs attribute in routed jobs

brianhlin:
TotalSubmitProcs is used in 8.5.7 for the for job totals in the new
condor_q formatting. If a user queues multiple jobs in their submit
file (resulting in multiple ProcIDs), the job router creates unique
ClusterIDs in the target SchedD for each ProcID. Each of these routed
jobs retains the TotalSubmitProcs from the original jobs and condor_q
reports erroneously large job totals.

@IainSteers encountered this issue and it's still being investigated since the target SchedD, if it's 8.5.7+, should be squashing the routed job's `TotalSubmitProcs` attribute. That being said, this change is defensive in nature and jobs without `TotalSubmitProcs` defined are unaffected.


djw8605:
Looks good to me

-Derek

PR #129, opened 2016-11-29
Accept Russian Data Intensive Grid certs

brianhlin:
GOC ticket here: https://ticket.opensciencegrid.org/31952 

We should think about adding something to the documentation so that users can identify the host cert not being in the map file as an issue and how to add a regex to the condor_mapfile themselves.


djw8605:
I hate that we have to do this for every different certificate.  @bbockelm , is there a better way?

brianhlin:
We actually have a ticket [here](https://jira.opensciencegrid.org/browse/SOFTWARE-2285) but it hasn't gone anywhere. Idk if we've got a technological solution so perhaps we can solve this through some documentation (or at least have something we can point users to).

brianhlin:
Oops, this fell by the wayside. Any objections to merging this?

djw8605:
go for it!

-Derek

PR #130, opened 2017-01-13
condor_ce_info_status: safely handle bad data

matyasselmeci:
Don't crash if entries are received from the collector with missing values for `CPUs` or `Memory`. Also handle if the attribute for `CPUs` is actually `Cpus`. This is for SOFTWARE-2185.

### DIFF
--- a/src/condor_ce_info_status
+++ b/src/condor_ce_info_status
@@ -14,6 +14,17 @@ import htcondor
 
 import htcondorce.info_query
 
+# Add htcondor.param.get() to old versions of condor-python that don't have it
+try:
+    htcondor.param.get
+except AttributeError:
+    def _htcondor_param_get(key, default=None):
+        try:
+            return htcondor.param[key]
+        except KeyError:
+            return default
+    htcondor.param.get = _htcondor_param_get
+
 
 def formatResourceAdsTable(resources, width):
     """Create a table of resource information obtained from `resources`,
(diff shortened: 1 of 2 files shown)


src/condor_ce_info_status
```diff
@@ -14,6 +14,17 @@ import htcondor
 
 import htcondorce.info_query
 
+# Add htcondor.param.get() to old versions of condor-python that don't have it
+try:
+    htcondor.param.get
```
bbockelm: When was `param.get` introduced - and what's the minimum version of HTCondor supported by the CE?

I suspect we can simply drop the workaround.

src/condor_ce_info_status
```diff
@@ -44,9 +55,9 @@ def formatResourceAdsTable(resources, width):
     table += "\n"
     for resource in resources:
         catalog_entry = resource.catalog_entry
-        table += "\n" + format_str % (catalog_entry['Name'],
-                                      catalog_entry['CPUs'],
-                                      catalog_entry['Memory'],
+        table += "\n" + format_str % (catalog_entry.get('Name','???'),
+                                      catalog_entry.get('CPUs', catalog_entry.get('Cpus', '???')),
```
bbockelm: Is `catalog_entry` a ClassAd object?  If so, lookup should be case-insensitive.

src/condor_ce_info_status
matyasselmeci: Yeah, I think it was for 8.2 compatibility. I'll drop it.

src/condor_ce_info_status
matyasselmeci: It should be in theory, but isn't in practice. I mentioned this in the ticket.

PR #132, opened 2017-02-03
Do not catch eval_set_ in JOB_ROUTER_ENTRIES (SOFTWARE-2581)

brianhlin:
We want to catch set_ATTR lines in JOB_ROUTER_ENTRIES for ATTR that
are eval_set_ATTR in JOB_ROUTER_DEFAULTS because they end up as
NOOPs. We were accidentally catching eval_set_ATTR statements in
JOB_ROUTER_ENTRIES, which we recommend against because they overwrite
their respective expressions in JOB_ROUTER_DEFAULTS but are
ultimately allowed.

### DIFF
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -45,7 +45,7 @@ for entry in JOB_ROUTER_CONFIG['JOB_ROUTER_ENTRIES']:
             + " Use the 'eval_set_' prefix instead."
 
     # Ensure that users don't set the job environment in the Job Router
-    if [x for x in entry.keys() if x.endswith('environment')]:
+    if any(x.endswith('environment') for x in entry.keys()):
         sys.exit("ERROR: Do not use the Job Router to set the environment. Place variables under " +\
                  "[Local Settings] in /etc/osg/config.d/40-localsettings.ini")
 
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -16,15 +16,16 @@ except ImportError:
              "Please ensure that the 'htcondor' and 'classad' are in your PYTHONPATH")
 
 # Create dict whose values are lists of ads specified in the relevant JOB_ROUTER_* variables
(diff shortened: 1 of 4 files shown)


brianhlin:
I think the output from `condor_config_val` will remain fairly stable but I'll check with TJ to see if we can rely on that.  Using the python bindings may be better here if possible (previous checks didn't work with them because they didn't handle multiple entries well) so I'll look into that, too.

brianhlin:
@djw8605 I've moved over all the CE config checks to a python script, which should give us better stability than parsing the condor tool outputs. Could you give this another review? Thanks

src/verify_ce_config.py
```diff
@@ -0,0 +1,64 @@
...
+try:
+    import classad
+    import htcondor
+except ImportError:
+    sys.exit("ERROR: Could not load HTCondor Python bindings. " + \
+             "Please ensure that the 'htcondor' and 'classad' are in your PYTHONPATH")
+
+# Create dict whose values are lists of ads specified in the relevant JOB_ROUTER_* variables
+JOB_ROUTER_CONFIG = dict((x, [x for x in classad.parseAds(htcondor.param[x])])
+                         for x in ['JOB_ROUTER_DEFAULTS', 'JOB_ROUTER_ENTRIES'])
```
djw8605: This formatting makes it difficult to understand what is going on here.

brianhlin: Haha yea, this is the one spot that I thought I was probably getting a little too carried away. I'll make this clearer.

djw8605: Well, if I was picky, you get more carried away later with if's inside the comprehension.

edquist: Sorry, a trigger word got my attention.

Is this `[x for x in classad.parseAds(htcondor.param[x])]`
any different from `list(classad.parseAds(htcondor.param[x]))` ?

brianhlin: @edquist No there isn't, fixed.

@djw8605 I think the `ifs` in list comprehensions are what makes them so useful :)

PR #139, opened 2017-03-03
Do not explicitly disable VOMS credential check.

bbockelm:
Instead, we'll depend on the default value that is compiled in to LCMAPS.  SOFTWARE-2633

### DIFF
--- a/src/condor_ce_env_bootstrap
+++ b/src/condor_ce_env_bootstrap
@@ -2,7 +2,6 @@
 
 # Configuration for LCMAPS
 export LLGT_LOG_IDENT=htcondor-ce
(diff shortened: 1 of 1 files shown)


bbockelm:
I'm not sure why we explicit disable validation in the environment variable: it's disabled by default in the OSG copy of `lcmaps`.

IIRC, the environment was taken from the Globus GridFTP environment at the time.  Maybe that environment file predated the change to LCMAPS?  Then, when we undid this setting on GridFTP, no one did the analogous action on this file?

Anyhow, we'll control this centrally via the `lcmaps` libraries so we can make the transition to enable validation simpler.

PR #141, opened 2017-04-13
Update central collectors via TCP

brianhlin:
UPDATE_COLLECTOR_WITH_TCP only controlled how the local collector is updated


bbockelm:
Are we sure we want this change?

IIUC, this means that the local collector may block when forwarding data to the centra collector.

If the CE collector develops problems, then this will filter down to the local collector, then to the schedd.  That creates a pretty tight tie between the central collector and the functionality of the CE itself.

brianhlin:
Correct, the local collector will block 20s for each collector in `CONDOR_VIEW_HOST`. How important is it to have ads in the central collector now that we're using it as an information service? With UDP, I can imagine only one of `collector1` and `collector2` getting the CE ad and running `condor_ce_info_status` and getting load balanced to the collector without the update

matyasselmeci:
Todd looked it up; the timeout for the TCP connection is 20. Can we live with that?

bbockelm:
The timeout for TCP connection is 20s, but how often will it be retried?  If it blocks for 20s, sleeps for 1ms, then blocks for 20s, ad infinitum, then the CE will still be hosed.

brianhlin:
Just spoke with Todd and it turns out that condor retries the TCP connection until success with some back off depending on the load it's creating on the collector. I created a [gittrac ticket](https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6249) so I'll close this for now.

PR #144, opened 2017-05-16
Fix bugs to pass GLUE2 Validator.

geonmo:
1. Remove unnecessary ".ch".
2. Remove disallowed "GLUE2COmputingShareMax*" on ResourceID.
3.  Add replace function to remove double quote.


bbockelm:
The approach looks fine to me.  Would you have some additional effort / time to include a unit test for Travis-CI?

geonmo:
@bbockelm Hi, could you give some information or other examples about Travis-CI?

bbockelm:
Hm - if you're not familiar with it, it might not be worth your time to do this as part of this PR.

Let me merge this one for now and we can circle back later.

PR #142, opened 2017-04-18
Add WholeNodeWanted ClassAd expression to request whole-node jobs

jthiltges:
When WholeNodeWanted=true, request the following resources for the job:
- TotalCpus of the worker
- 95% of TotalMemory

Implements the method from "Getting To Multicore: Resizing jobs in CMS" by @bbockelm. HTCondor Week 2016.
https://research.cs.wisc.edu/htcondor/HTCondorWeek2016/presentations/ThuBockelman_ResizableJobs.pdf

PR #145, opened 2017-05-16
Use WantWholeNode to match existing factory naming convention

jthiltges:
OSG factory operations has an existing convention of "Want{Item}", e.g. WantSingularity.

For whole node support added in #142, change flag from WholeNodeWanted to WantWholeNode to match existing factory naming convention.


jthiltges:
Thank you for your help, Brian!

PR #146, opened 2017-05-25
Bump version to 2.2.0

brianhlin:
Prepping for the next OSG release. Is there anything else we want to sneak in before the dev freeze Tuesday?


djw8605:
I can't think of anything else to add.

PR #150, opened 2017-06-13
Split out non-OSG bits into htcondor-ce-core* packages

brianhlin:
First pass at separating the OSG bits so non-OSG sites can more readily use the HTCondor-CE as their gateway solution. Eventually, we'd like to get the configuration into HTCondor itself as a `CE` or `gateway` configuration template.


config/htcondor-ce.spec
```diff
@@ -76,6 +75,21 @@ Requires: /usr/bin/unshare
 %description
 %{summary}
 
+%package core
+Group: Applications/System
+Summary: Core configuration for HTCondor-CE
+
+Provides:  condor-ce = %{version}
+Provides:  %{name}-master = %{version}-%{release}
```
brianhlin: I'm not sure if these `Provides:` should go into the core or the base package.

config/condor_mapfile
```diff
@@ -1,10 +1,3 @@
-GSI "^\/DC\=com\/DC\=DigiCert-Grid\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
-GSI "^\/DC\=DigiCert-Grid\/DC\=com\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
-GSI "^\/DC\=org\/DC\=opensciencegrid\/O=Open Science Grid\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
-GSI "^\/C\=RU\/O\=RDIG\/OU\=hosts\/OU=*\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
-GSI "^\/C\=BR\/O\=ANSP\/OU\=ANSPGrid\ CA\/OU\=Services\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
-GSI "^\/DC\=org\/DC\=terena\/DC\=tcs.*\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@daemon.opensciencegrid.org
-GSI "^\/DC\=ch\/DC\=cern\/OU\=computers\/CN\=(host\/)?([A-Za-z0-9.\-]*)$" \2@cern.ch
 GSI (.*) GSS_ASSIST_GRIDMAP
 GSI "(/CN=[-.A-Za-z0-9/= ]+)" \1@unmapped.opensciencegrid.org
```
djw8605: Should this still be opensciencegrid.org ?

config/htcondor-ce.spec
```diff
@@ -418,16 +445,21 @@ fi
 %{_datadir}/condor-ce/config.d/02-ce-bosco-defaults.conf
 
 %files client
+%config(noreplace) %{_sysconfdir}/condor-ce/condor_mapfile.osg
+%config(noreplace) %{_sysconfdir}/condor-ce/config.d/01-osg.conf
```
djw8605: Why is this mapfile and config going in the client?  Isn't it needed in the actual CE?

config/condor_mapfile
brianhlin: That's one other thing I was debating: on one hand it's just a cosmetic change, on the other it probably needs to be renamed if we want to push config upstream. If we change it in `condor_mapfile`, we should also change it `condor_mapfile.osg` so we can keep 01-ce-auth.conf the same.

config/htcondor-ce.spec
brianhlin: The original mapfile was in the client as well, though I'm not entirely sure why. The actual CE pulls in the client by default so it's not like you'd be able to run a CE without the mapfile.

bbockelm: The mapfile needs to be in the client because the client wants to authenticate the remote CE as well.

config/condor_mapfile
bbockelm: Yeah - you want to keep it as `@opensciencegrid.org`.  The name itself is arbitrary, but needs to be consistent across all CEs.

bbockelm:
@matyasselmeci - can you review this in terms of any issues with `osg-configure`?

matyasselmeci:
Unless we're changing which packages our OSG users install to get a CE, this has no effects on osg-configure.

brianhlin:
Do we really need the original mapfile (with all the OSG DNs) in the base client? Doesn't the [OSG-less mapfile](https://github.com/brianhlin/htcondor-ce/blob/b1958fb891d9fa4afff75a22236e632727f9464c/config/condor_mapfile) suffice with perhaps some documentation on adding DNs as needed?

brianhlin:
@matyasselmeci adding you for your opinion on packaging

config/htcondor-ce.spec
matyasselmeci: Depends on what requires them. If whatever requires them will not work without the OSG bits too, then the Provides should go in the main package.

brianhlin: @matyasselmeci ^^

config/htcondor-ce.spec
```diff
@@ -58,6 +57,21 @@ Requires: /usr/bin/unshare
...
+Group: Applications/System
+Summary: Core configuration for HTCondor-CE
+
+Provides:  condor-ce = %{version}
+Provides:  %{name}-master = %{version}-%{release}
+
+Requires: condor
+Requires: %{name}-core-client = %{version}-%{release}
+
+Obsoletes: condor-ce < 0.5.4
```
matyasselmeci: Again, you should be careful with this so you don't replace condor-ce with htcondor-ce-core for installations that really need the OSG bits.

config/htcondor-ce.spec
```diff
@@ -167,6 +181,19 @@ Provides:  condor-ce-bosco = %{version}
...
+Provides:  condor-ce-client = %{version}
+
+%description client
+%{summary}
+
+%package core-client
+Group: Applications/System
+Summary: Client-side tools for submission to HTCondor-CE
+
+BuildRequires: boost-devel
```
matyasselmeci: All BuildRequires should go in the main package.

config/condor_mapfile
brianhlin: Why does it have to be consistent across all CEs? Isn't the authz all internal to each CE?

config/htcondor-ce.spec
brianhlin: The `Obsoletes`/`Provides` are from 2012 and nothing in our 3.4 repos require `condor-ce-*` so I think we can probably drop them entirely - at the same time I don't think there's any harm in keeping them around

brianhlin:
@matyasselmeci ready for re-review. I dropped the provides/obsoletes for `condor-ce*` completely because they were used for the package name transition in 2012. Perhaps we could do the same thing for the other packages since nothing requires `condor-ce*` in 3.4

matyasselmeci:
Our unofficial policy is that if the Obsoletes/Provides is required for a clean upgrade from 3.3, then we still need 'em, but if it's only needed for a clean upgrade from 3.2 or 3.1, then it's not needed.

config/htcondor-ce.spec
brianhlin: If we can choose an arbitrary suffix other than `@opensciencegrid.org`, I think we could revert this change and have a single `condor_mapfile`. That being said, we should probably ship a `condor_config.template` and provide a config(noreplace) `condor_mapfile` if we really want to make this non-OSG-friendly.

brianhlin:
Remaining work to be done:

- Separate the `BLAHP_LOCATION` into a separate config file so OSG builds will use the OSG blahp and the UW build uses the built-in blahp. Keep in mind that these two versions of the blahp 
- Change the `*.opensciencegrid.org` authz config everywhere? This may not be immediately possible because `01-ce-auth.conf` is `%config(noreplace)` and `condor_mapfile` is not, so if we update the authz suffix then OSG users will need to manually update their config to match our new suffix. 
- ...?

brianhlin:
Still need to:

- [ ] Document how to map VOMS attr using the `condor_mapfile`
- [ ] Test blahp integration with both builds

brianhlin:
@matyasselmeci @bbockelm @djw8605 ready for re-review

CMakeLists.txt
```diff
@@ -38,14 +38,104 @@ configure_file (
...
-install(PROGRAMS src/condor_ce_startup src/verify_ce_config.py src/condor_ce_startup_internal src/condor_ce_env_bootstrap src/condor_ce_client_env_bootstrap src/condor_ce_router_defaults src/local-wrapper src/condor_ce_jobmetrics src/condor_ce_metric src/condor_ce_view src/bdii/htcondor-ce-provider src/gratia_cleanup.py DESTINATION ${SHARE_INSTALL_PREFIX}/condor-ce)
-install(PROGRAMS src/condor_ce_config_generator src/condor_ce_config_val src/condor_ce_history src/condor_ce_hold src/condor_ce_info_status src/condor_ce_q src/condor_ce_qedit src/condor_ce_release src/condor_ce_rm src/condor_ce_submit src/condor_ce_version src/condor_ce_reconfig src/condor_ce_router_q src/condor_ce_status src/condor_ce_reschedule src/condor_ce_run src/condor_ce_trace src/condor_ce_ping src/condor_ce_off src/condor_ce_on src/condor_ce_restart src/condor_ce_job_router_info src/condor_ce_host_network_check DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
-install(FILES src/htcondorce/__init__.py src/htcondorce/web_utils.py src/htcondorce/web.py src/htcondorce/rrd.py src/htcondorce/tools.py src/htcondorce/info_query.py src/audit_payloads.py DESTINATION ${PYTHON_SITELIB}/htcondorce)
+install(PROGRAMS
```
matyasselmeci: THANK YOU

brianhlin:
@djw8605 nothing that I saw

PR #151, opened 2017-06-15
Save the number of CPUs allocated at last match

djw8605:
When a whole node job is matched, we need to save the number of cpus
so that accounting systems can pick up how many cpus where actually 
used by a job.


brianhlin:
Since it's saved as a string, is the plan to convert it to an int on the GRACC end?

djw8605:
Probably convert it to an in on the gratia-probe side.

bbockelm:
Can we put in a PR for the corresponding Gratia probe changes (if needed?).  Once that's done, I will approve this PR.

djw8605:
@bbockelm Sounds good.  does this pull request look good though?  I was wondering if there was a use for this attribute `set_JOB_GLIDEIN_Cpus` ?  Since we are overwriting it for Gratia's own use.

djw8605:
@bbockelm is out this week, but we have merged https://github.com/opensciencegrid/gratia-probe/pull/18

@brianhlin can you check this pull request, and merge it?

PR #152, opened 2017-06-20
Pass OriginalMemory variable through GAHP

jthiltges:
Hi @djw8605, @brianhlin, and @bbockelm,

Troubleshooting failed jobs, I found that the memory requirements were not passing through BLAHP to the local scheduler (pbs_submit.sh -> SLURM).

When passing a job to a GAHP server, gridmanager [copies](https://github.com/htcondor/htcondor/blob/f7ac6cce/src/condor_gridmanager/infnbatchjob.cpp#L1381-L1396) a limited list of attributes, which includes RequestMemory.

With the whole node changes (#142), RequestMemory is now an expression that includes OriginalMemory. But, OriginalMemory is not passed along by gridmanager, so BLAHP cannot evaluate RequestMemory.

Please consider this PR as an ugly proof-of-concept. It appears to get RequestMemory passed to the local scheduler, but I'd appreciate your help on making it cleaner.

Regards,
John

### DIFF
--- a/src/condor_ce_router_defaults
+++ b/src/condor_ce_router_defaults
@@ -51,6 +51,8 @@ JOB_ROUTER_DEFAULTS_GENERATED = \\
     /* Note default memory request of 2GB */ \\
     /* Note yet another nested condition allow pass attributes (maxMemory,xcount,jobtype,queue) via gWMS Factory described within ClassAd */ \\
     eval_set_OriginalMemory = ifThenElse(maxMemory isnt undefined, maxMemory, ifThenElse(default_maxMemory isnt undefined, default_maxMemory, 2000)); \\
+    /* Duplicate OriginalMemory expression and add remote_ prefix. This passes the attribute from gridmanager to BLAHP. */ \\
+    eval_set_remote_OriginalMemory = ifThenElse(maxMemory isnt undefined, maxMemory, ifThenElse(default_maxMemory isnt undefined, default_maxMemory, 2000)); \\
     set_JOB_GLIDEIN_Memory = "$$(TotalMemory:0)"; \\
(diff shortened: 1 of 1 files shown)


jthiltges:
Hi Brian. Thank you for taking a look. On first glance, I agree with your suggestion. Though I believe there are two paths to worry about:
* CE => Grid manager (slurm, pbs, lsf, sge, etc.)
* CE => Condor

The grid manager [removes](https://github.com/htcondor/htcondor/blob/f7ac6cce/src/condor_gridmanager/infnbatchjob.cpp#L1649) the "remote_" prefix from attributes, converting remote_X to X. But condor receives all the attributes without any mangling.

| Destination  | Input attribute       | Output attribute      |
| ------------ | --------------------- | --------------------- |
| Grid manager |        OriginalMemory |                _none_ |
| Grid manager | remote_OriginalMemory |        OriginalMemory |
| Condor       |        OriginalMemory |        OriginalMemory |
| Condor       | remote_OriginalMemory | remote_OriginalMemory |

We need the "OriginalMemory" attribute to be set in the output for both paths, grid manager and condor. Duplicating the expression seems ugly, but I think some sort of duplication is required.

jthiltges:
Hi Brian,

I updated the PR with a more detailed commit message, a comment in the job router, and revised it a bit.

It now just sets "remote_OriginalMemory = OriginalMemory". It's still not pretty, but removes the repetition of the OriginalMemory expression. From some quick testing on a gatekeeper, the memory requirements are passing through OK.

Regards,
John

jthiltges:
Well, bother. I'd hoped it would evaluate top-down. Thank you for pointing that out. PR updated.

PR #153, opened 2017-06-22
Add audit_payloads plugin

DrDaveD:


DrDaveD:
@brianhlin and @bbockelm please review.

config/01-ce-collector-audit-payload.conf
```diff
@@ -0,0 +1,24 @@
+# Settings for the collector audit payloads function.
+# The easiest way to direct the payload logging information here is to
+# set the environment variable CONDORCE_COLLECTOR_HOST on worker nodes
+# pointing to this condor-ce host.  Then pilot jobs that use condor
```
bbockelm: This is automatically done by the CE, so we don't need this comment - should be redundant.

config/01-ce-collector-audit-payload.conf
```diff
@@ -0,0 +1,24 @@
...
+# will automatically forward their payload classads here for auditing.
+
+# You can edit this file but that prevents future rpm upgrade
+# replacements, so it is better to override these settings in 
+# /etc/condor-ce/99-ce-collector-audit-payload.conf
```
bbockelm: If it goes into `/etc/condor-ce/config.d`, then the admins should be able to edit it.  Defaults should go into `/usr/share/condor-ce/config.d` and _not_ be marked as config files.

config/01-ce-collector-audit-payload.conf
```diff
@@ -0,0 +1,24 @@
...
+
+# Set this value to enable the plugin
+# COLLECTOR_PLUGIN_PYTHON_MODULES = htcondorce.audit_payloads
+
+# Set these values to enable the log file
+# COLLECTOR_AUDIT_LOG = $(LOG)/AuditPayloadLog
+# COLLECTOR_DEBUG = D_AUDIT, $(COLLECTOR_DEBUG)
```
bbockelm: Enable all of the above by default.

config/htcondor-ce.spec
```diff
@@ -477,6 +480,7 @@ fi
 %config(noreplace) %{_sysconfdir}/condor-ce/config.d/01-ce-collector.conf
 %config(noreplace) %{_sysconfdir}/condor-ce/config.d/01-ce-collector-requirements.conf
 %config(noreplace) %{_sysconfdir}/condor-ce/config.d/01-ce-auth.conf
+%config(noreplace) %{_sysconfdir}/condor-ce/config.d/01-ce-collector-audit-payload.conf
```
bbockelm: I'm not sure it should be in this sub-RPM?  I think you only want it in the above one.

config/htcondor-ce.spec
```diff
@@ -493,6 +497,9 @@ fi
 %attr(1777,root,root) %dir %{_localstatedir}/lib/gratia/condorce_data
 
 %changelog
+* Thu Jun 22 2017 Dave Dykstra <dwd@fnal.gov> - 2.2.1-1
+- Add audit_payloads function
```
bbockelm: Can you be more descriptive what this means for the admins?

src/audit_payloads.py
```diff
@@ -0,0 +1,194 @@
...
+# There is one condor-ce configuration variable AUDIT_PAYLOAD_MAX_HOURS,
+#   which is optional and indicates the maximum number of hours any job
+#   is expected to run, default 72 hours (3 days).  After that time the
+#   jobs will stop being tracked, in case a stop message was missed.
+#
+# Written by Dave Dykstra, June 2017
+#
+
+import htcondor
+# I would use the simpler "time" module but it gets a load error
```
bbockelm: That sounds a bit scary - can you hunt it down a bit?  Is there a version mismatch between your compiled version of python and the one condor was compiled against?

src/audit_payloads.py
```diff
@@ -0,0 +1,194 @@
...
+    # remove this job from the doubly-linked list
+    thisjob = runningjobs[idx]
+    if thisjob['previdx'] == None:
+	firstidx = thisjob['nextidx']
+    else:
+	runningjobs[thisjob['previdx']]['nextidx'] = thisjob['nextidx']
+    if thisjob['nextidx'] == None:
+	lastidx = thisjob['nextidx']
+    else:
+	runningjobs[thisjob['nextidx']]['previdx'] = thisjob['previdx']
```
bbockelm: I don't think any of this linked list stuff is necessary.  Remove it all, throughout the module.

src/audit_payloads.py
```diff
@@ -0,0 +1,194 @@
...
+    runningjobs[idx] = thisjob
+
+# a job may be being stopped
+def stopjob(info):
+    global runningjobs
+    if 'Name' not in info or 'SlotID' not in info:
+	return
+    idx = (info['Name'], info['SlotID'])
+    if idx not in runningjobs:
+	return
```
bbockelm: The above 5 lines seem to be mostly duplicated through the module.  Can you refactor them into a helper function?

src/audit_payloads.py
```diff
@@ -0,0 +1,194 @@
...
+
+    if maxjobsecs == None:
+	# initialize this global variable
+	if 'AUDIT_PAYLOAD_MAX_HOURS' in htcondor.param:
+	    maxjobhours = int(htcondor.param['AUDIT_PAYLOAD_MAX_HOURS'])
+	else:
+	    maxjobhours = 3 * 24
+	htcondor.log(htcondor.LogLevel.Audit,
+	    "Audit payload maximum job hours: %d" % maxjobhours)
+	maxjobsecs = maxjobhours * 60 * 60
```
bbockelm: Rather than caching it on first call, can you simply do this at the module level when things are loaded?

src/audit_payloads.py
```diff
@@ -0,0 +1,194 @@
...
+	    break
+	nextidx = thisjob['nextidx']
+	loginfo = {}
+	loginfo['Name'] = idx[0]
+	loginfo['SlotID'] = idx[1]
+	loginfo['GlobalJobId'] = thisjob['globaljobid']
+	htcondor.log(htcondor.LogLevel.Audit,
+	    "Cleaning up %d-second expired job: %s" % (deltasecs, loginfo))
+	removerunningjob(idx)
+	idx = nextidx
```
bbockelm: Two thoughts:
* Just iterate through the `runningjobs` array using the built-in iterators (`for idx, thisjob in runningjobs.items():`).
* Seems costly to do this every update, given that there might be many updates on a busy CE.  Maybe once every N seconds?  (N=60?)

config/01-ce-collector-audit-payload.conf
DrDaveD: Oh wow, that's a surprise to me.  What were all those classads getting sent to the condorce collector for before this time?

config/01-ce-collector-audit-payload.conf
DrDaveD: I understand, and this file is editable, but my comment is true because editing it means that the next upgrade will result in a .rpmnew and a little more hassle than putting changes into a new file.  I did forget the config.d in the path though, I'll add that.

config/htcondor-ce.spec
DrDaveD: I wasn't sure.  What is the htcondor-ce-collector rpm for?  It already had other files that were in the main htcondor-ce so I assumed it was an rpm if somebody wanted to alternately install only the collector.

config/01-ce-collector-audit-payload.conf
DrDaveD: Ok.  I thought this was going to be opt-in.

DrDaveD: Wait, by all of the above did you mean to include setting COLLECTOR_PLUGIN_PYTHON_MODULES = htcondorce.audit_payloads to enable the plugin by default, or just enable the log file?

src/audit_payloads.py
DrDaveD: I don't understand.  Don't I have to keep track of expired entries?  I don't know how to do that efficiently without the linked list.

src/audit_payloads.py
DrDaveD: I did try to hunt it down, and it was a common problem and a known bug in python.  The other workarounds were all more onorous than his one.  See for example this [stack overflow question](https://stackoverflow.com/questions/11842920/undefined-symbol-pyexc-importerror-when-embedding-python-in-c).  It includes an answer that refers to a cilogon.org page.

src/audit_payloads.py
DrDaveD: I only saw two places.  I tried it anyway, but the second usage in startjob() doesn't actually save any lines of code so I don't really see an advantage.

src/audit_payloads.py
DrDaveD: The list is sorted so it is not very costly.  All it does every time is compare the at time of the oldest one to the expiration time. I don't see that as being much slower than comparing the current time to 60 seconds ago, is it?

src/audit_payloads.py
DrDaveD: I wasn't sure but yes you're right. I'll do that.

src/audit_payloads.py
DrDaveD: Judging by your comment below, maybe you didn't realize that using the links enables keeping the jobs sorted by age.   New jobs are added to the end of the list, and old jobs are checked at the beginning of the list.  The double links allows removing jobs from the middle of the list when they stop.  I think this is much more efficient than searching through all the jobs once a minute, making everything order(1) instead of order(n), especially as n gets large.

config/htcondor-ce.spec
brianhlin: The collector RPM is for the OSG central collector so this file isn't necessary

DrDaveD: Ok, I removed it.

src/audit_payloads.py
DrDaveD: I added a comment to explain the reason for this.

DrDaveD:
@bbockelm can you please re-review?

src/audit_payloads.py
brianhlin: I see what you're doing Dave but why not use `collections.OrderedDict` instead? It'd be much cleaner.

Since it's only available starting in `python-2.6.6-43` (RHEL 6.5), we could just start shipping the OrderedDict class.

config/htcondor-ce.spec
```diff
@@ -2,7 +2,7 @@
 #define gitrev osg
 
 Name: htcondor-ce
-Version: 2.2.0
+Version: 2.2.1
```
DrDaveD: done

src/audit_payloads.py
```diff
@@ -0,0 +1,195 @@
...
+if 'AUDIT_PAYLOAD_MAX_HOURS' in htcondor.param:
+    maxjobhours = int(htcondor.param['AUDIT_PAYLOAD_MAX_HOURS'])
+else:
+    maxjobhours = 3 * 24
+htcondor.log(htcondor.LogLevel.Audit,
+    "Audit payload maximum job hours: %d" % maxjobhours)
+maxjobsecs = maxjobhours * 60 * 60
+
+def dumpvars():
+    htcondor.log(htcondor.LogLevel.Audit, "firstidx: %s, lastidx: %s, runningjobs: %s" % (firstidx, lastidx, str(runningjobs)))
```
DrDaveD: Sure, done

src/audit_payloads.py
DrDaveD: I was unaware of OrderedDict.  I see it does the same thing with a doubly-linked list.  I will switch to it, and include this [backported copy](https://code.activestate.com/recipes/576693/) for ```%if 0%{?rhel} < 7```

src/audit_payloads.py
bbockelm: Then this is a HTCondor bug - HTCondor should be explicitly linking against `libpython` on appropriate platforms.

@brianhlin - can you report upstream?

bbockelm:
Checking over this PR again -- seems we are waiting for the switch + backport of`OrderedDict`, right?

DrDaveD:
That's correct, we're waiting on me to switch to OrderedDict.  Since we missed the July freeze deadline I put it off until after my 4th of July weekend vacation.

bbockelm:
Sounds good - I just wanted to double-check I had read the ticket status correctly.

src/audit_payloads.py
brianhlin: @DrDaveD what version of condor did you see this problem with? Does the following reproduce the issue?

```
import htcondor
import time
```

DrDaveD: It was 8.6.3.  All it takes to reproduce is "import time".  When I do that I see this in CollectorLog if D_FULLDEBUG is enabled:
```
07/07/17 12:14:10 Successfully loaded plugin: /usr/libexec/condor/libcollector_python_plugin.so
07/07/17 12:14:10 Python exception occurred when loading module htcondorce.audit_payloads: Traceback (most recent call last):

  File "/usr/lib/python2.7/site-packages/htcondorce/audit_payloads.py", line 1, in <module>
    import time

ImportError: /usr/lib64/python2.7/lib-dynload/timemodule.so: undefined symbol: PyExc_ValueError
```

bbockelm: What happens when you do:

```
ldd /usr/libexec/condor/libcollector_python_plugin.so
```

?  Do you see it explicitly linked against `libpython2.7.so.1.0` (python 2.6 on RHEL6, of course)?  I see that symbol:

```
$ nm -D  /lib64/libpython2.7.so.1.0 | grep PyExc_ValueError
00000000003898e8 D PyExc_ValueError
```

DrDaveD: Yes I see both of those things.  My testing machine is RHEL7.3.

bbockelm: Ok: @brianhlin - can you open a HTCondor ticket?  I don't want to go too far on a tangent here chasing what might be a HTCondor bug.

DrDaveD: Uh oh now I've got a similar error with OrderedDict:
```
ImportError: /usr/lib64/python2.7/lib-dynload/_collectionsmodule.so: undefined symbol: _Py_ZeroStruct
```
Googling for that error got me to this [stackoverflow question](https://stackoverflow.com/questions/8302810/undefined-symbol-in-c-when-loading-a-python-shared-library) which suggests that the dlopen that loads the python should add the RTLD_GLOBAL option.

I plan to see if I can add that to the dlopen() in htcondor/src/condor_utils/LoadPlugins.cpp

DrDaveD: Yes, that worked.

DrDaveD:
Ok here's a new version based on the OrderedDict and time modules. @bbockelm and @brianhlin please re-review.

I went through considerable effort to add the code for backporting OrderedDict to EL6, but when I got up to testing it I found that OrderedDict was already there.  Redhat must have backported it already, it is in the base SL6 install.

DrDaveD:
The travis check failed because condor >= 8.6.4-1.1 does not exist yet.

config/01-ce-collector-audit-payload.conf
```diff
@@ -0,0 +1,20 @@
+# Settings for the collector audit payloads function.
```
bbockelm: @brianhlin - we tend to have config files in `/etc/condor-ce/config.d` for two reasons:
1.  Historical: they were here before they were in `/usr/share/condor-ce` and we wanted some continuity.
2.  We want to communicate / highlight some config setting the admin may want to change.

Does either apply here?  The only potential tunable is `AUDIT_PAYLOAD_MAX_HOURS`, but I suspect that the admin would have to be such a deep expert to want to tune that they don't need the help text.

I'm tempted to just suggest removing this file completely; seems it would make things simpler.

bbockelm:
If you ping this ticket when 8.6.4-1.1 exists in `osg-development`, I can retrigger the Travis-CI tests.

config/01-ce-collector-audit-payload.conf
DrDaveD: I can't drop the file, but do you mean I should move it to /usr/share/condor-ce/config.d (and change the comments accordingly)?

brianhlin: Yup, sorry, let's move it to `/usr/share/condor-ce/config.d/`

brianhlin:
The EL6 errors are troubling so I've kicked off VMU tests against 8.6.4-1.1. I've also kicked off [Travis-CI tests](https://travis-ci.org/brianhlin/htcondor-ce/jobs/252980247) with a modified osg-test that should timeout the `condor_ce_run` command after 5 min so we can avoid the Travis-CI timeout and get condor logs back.

brianhlin:
VMU tests look good: http://vdt.cs.wisc.edu/tests/20170712-1604/results.html with condor-8.6.4-1.1 and the release version of htcondor-ce

The travis-ci tests didn't fare any better (looks like the osg-test timeout code doesn't work). @DrDaveD can you set up an EL6 CE using your collector plugin and try validating it manually?

DrDaveD:
Ok I moved the default file, and successfully tested this with my own build of htcondor-8.6.4-1.1 on both EL6 and EL7.  Unfortunately it does not work the OSG build of htcondor-8.6.4-1.1, on either O.S., and I don't know why yet.  The CollectorLog shows it loading /usr/libexec/condor/libcollector_python_plugin.so, so it's not the problem we had with 8.6.3, but it doesn't run the audit_payloads plugin nor print any error from it.  My build came directly from htcondor's build-on-linux.sh

bbockelm:
Once we figure out why the unit tests are failing, I'm ready for this to go.

DrDaveD:
I haven't figured it out exactly, but I can reproduce the bad behavior of condor not running my plugin if I do "rpmbuild --rebuild --define 'osg 1' condor-8.6.4*src.rpm".  Unfortunately that makes a very big difference in the build flags, including undefining the uw_build flag, and I don't know which part makes the difference.  strace hasn't been much help, and I haven't yet figured out how to run it under gdb.

timtheisen:
I am thinking that it would be helpful to stage the prerelease for 8.6.5 in testing repository for you to give a build done on the UW build machine a spin. Let me know if you would find that helpful.

DrDaveD:
@timtheisen Thanks but I don't think it will help since the problem is only seen with --define 'osg 1'.  Unless you know of another bug fix in 8.6.5 that might be relevant?  It doesn't have to do with the patch that adds RTLD_GLOBAL.  I do not see the problem in 8.6.3-1.1.osg34.

DrDaveD:
I have updated the pull request to be able to handle job stops from glideinWMS partitionable slots.  It is unfortunately considerably more complicated because sometimes there aren't stop messages received for the individual slots.  It has to parse the name to look for "slotN@X" which can indicate to stop all jobs with the corresponding names "slotN_M@X".

I also updated the spec file to require condor >= 8.6.5.

bbockelm:
Can we get 8.6.5 into the osg-development repo so the CI tests can run?

brianhlin:
ETA for 8.6.5 test build is Monday

brianhlin:
EL6 failures due to unprivileged docker run, fixed on my branch: https://travis-ci.org/brianhlin/htcondor-ce/builds/257781750

I believe @DrDaveD's out on vacation so we can add my commit after merging this.

PR #158, opened 2017-07-20
Only complain about not being osg-configured if osg-configure is present

matyasselmeci:
(SOFTWARE-2805)


matyasselmeci:
Tried this out on an SLF7 VM and got the expected results based on the presence of 50-osg-configured.conf and 50-osg-configure.conf

matyasselmeci:
My bad, it's 50-osg-configure.conf and 50-osg-configure-present.conf.

brianhlin:
Ok, I can live with that.

PR #164, opened 2017-07-31
fix when a startjob first stops the last job in a master

DrDaveD:
This fixes a bug I found that was losing a small percentage of stop messages.

### DIFF
--- a/config/htcondor-ce.spec
+++ b/config/htcondor-ce.spec
@@ -2,8 +2,8 @@
 #define gitrev osg
 
 Name: htcondor-ce
-Version: 3.0.0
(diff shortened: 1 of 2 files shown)


config/htcondor-ce.spec
```diff
@@ -3,7 +3,7 @@
 
 Name: htcondor-ce
 Version: 3.0.0
-Release: 1%{?gitrev:.%{gitrev}git}%{?dist}
+Release: 2%{?gitrev:.%{gitrev}git}%{?dist}
```
brianhlin: Could you instead bump the version to `3.0.1`?

DrDaveD: Yes, sorry, I should have thought of that.  Done.

PR #165, opened 2017-08-09
Pass on gwms distro requests to condor backends (SOFTWARE-2801)

brianhlin:
As requested by @jdost321

### DIFF
--- a/src/condor_ce_router_defaults
+++ b/src/condor_ce_router_defaults
@@ -47,7 +47,9 @@ JOB_ROUTER_DEFAULTS_GENERATED = \\
     \\
     /* Set new requirements */ \\
     /* set_requirements = LastClientContact - time() < 30*60; */ \\
-    set_requirements = True; \\
+    /* Respect gwms factory distro requests (RHEL6/RHEL7 are used to request EL6/EL7) */ \\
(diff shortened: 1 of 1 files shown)


djw8605:
Though, I'm open to a discussion.

brianhlin:
We could add a config knob that controls whether or not we use this requirements expression.

djw8605:
Really, I would like @bbockelm to weigh in.  I may be just blowing our use case out of proportion.

bbockelm:
- What happens when a glidein requests `any`?
- What happens when there is a request for a non-RHEL site?  (i.e., Debian)?  I think we should probably whitelist known distros instead of blindly applying `substr`.
- Maybe have this requirement be an `eval_set` and be disabled if `WantDocker` is true?
- I would prefer to see the factory use the HTCondor nomenclature.

brianhlin:
@bbockelm so you envision that the factories would submit pilots with something like the following?

```
requirements = "OpSysMajorVer =?= '6'"
```

brianhlin:
I was going through old JIRA tickets and was reminded of this PR. If this is something we still want, I'm still unclear about what you meant here @bbockelm :

> I would prefer to see the factory use the HTCondor nomenclature.

brianhlin:
Abandoning this

PR #172, opened 2017-10-09
fix comment -- remove explicit port for CE

DrDaveD:
I found when testing this through condor_ce_run that the setting for $CONDORCE_COLLECTOR_HOST includes the port :9619, so I remove the port in the comments about how to use the audit payload logging function.


DrDaveD:
Ok, I updated it.

PR #173, opened 2017-11-16
Update help e-mail (SOFTWARE-3014)

brianhlin:


matyasselmeci:
or you can merge #173 which looks exactly the same as this.

PR #175, opened 2017-11-20
New provider

lfield:
This is a new version of the provider for the HTCondor CE.  Note that as the HTCondor CE is a new(ish) service, GLUE 1.3 publication is not required. This provider is a stripped down version that should suit the needs of Alice and LHCb.


src/bdii/htcondor-ce-provider
```diff
@@ -0,0 +1,225 @@
...
+        if not schedd_ad['AuthenticatedIdentity'] == 'computing-element@cern.ch':
+            continue
+
+        try:
+            if htcondor.SecMan().ping(schedd_ad, "READ")['AuthorizationSucceeded']:
+                state = 'ok'
+            else:
+                state = 'warning'
+            state_info = 'Authorization ping successful'
+        except Exception, e:
```
brianhlin: Can we narrow the scope of this exception? I only get `RuntimeError` when I'm not authorized for a specific command or I give it a bad ad. Also, nothing seems to be done with `e` so it can be dropped

src/bdii/htcondor-ce-provider
```diff
@@ -0,0 +1,225 @@
...
+                continue
+            total_vo_jobs[job['x509userproxyvoname']] += 1
+            if job['JobStatus'] == 1:
+                idle_vo_jobs[job['x509userproxyvoname']] += 1
+            elif job['JobStatus'] == 2:
+                running_vo_jobs[job['x509userproxyvoname']] += 1
+
+        cmd = ['/usr/bin/openssl', 'x509', '-noout', '-issuer', '-nameopt', 'RFC2253', '-in', '/etc/grid-security/hostcert.pem']
+        cp = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        issuer = cp.communicate()[0].replace('issuer= ','').strip()
```
brianhlin: Could we replace the `'issuer= '` with `r'issuer=\s*'`?

src/bdii/htcondor-ce-provider
```diff
@@ -0,0 +1,225 @@
...
+            print policy_ldif %{
+                'policyid': policyid,
+                'shareid': shareid, 
+                'bind_dn': bind_dn,
+                'vo': vo,
+                }
+
+if __name__ == '__main__':
+    main()
+
```
brianhlin: Let's remove the extra newlines

lfield:
I have addressed your comments and made the changes as suggested. I did not address the packaging. I anticipate that this provider should replace the existing one.

src/bdii/htcondor-ce-provider
```diff
@@ -0,0 +1,245 @@
...
+        if not schedd_ad['AuthenticatedIdentity'] == 'computing-element@cern.ch':
+            continue
+
+        try:
+            if htcondor.SecMan().ping(schedd_ad, "READ")['AuthorizationSucceeded']:
+                state = 'ok'
+            else:
+                state = 'warning'
+            state_info = 'Authorization ping successful'
+        except KeyError, e:
```
brianhlin: You should also catch `RuntimeError` if ping fails authorization

src/bdii/htcondor-ce-provider
```diff
@@ -0,0 +1,245 @@
...
+        'central_manager': central_manager, 
+        'bind_dn': bind_dn, 
+        'num_endpoints': len(schedd_ads),
+        'num_shares': len(vonames),
+        'site_name': site_name,
+        }
+
+    for schedd_ad in schedd_ads:
+
+        if not schedd_ad['AuthenticatedIdentity'] == 'computing-element@cern.ch':
```
kreczko: Also, I am reading this correctly? This provider will not work if `chedd_ad['AuthenticatedIdentity'] != 'computing-element@cern.ch'`?

src/bdii/htcondor-ce-provider
```diff
@@ -0,0 +1,245 @@
...
+    print resource_ldif %{
+        'central_manager': central_manager, 
+        'bind_dn': bind_dn,
+        'instances': instances,
+        }
+
+    schedd_ads = coll.locateAll(htcondor.DaemonTypes.Schedd)
+
+    # Print the entry for the GLUE2 Service
+    print service_ldif %{
```
kreczko: Since Python 2 is pretty much going into legacy mode, wouldn't it make sense to `from __future__ import print_function` and use `"{name}".format(name=value)` instead of `"%(name)" % {name:value}`?

kreczko:
Apologies for my ignorance, but is there a connection to https://github.com/opensciencegrid/htcondor-ce/blob/master/src/bdii/condor_ce_bdii_generate_glue2.py ? Many lines seem very similar, especially the templates.

lfield:
I have addressed all the comments. Yes this provider is an evolution of condor_ce_bdii_generate_glue2.py. Note that GLUE 1.3 is not required for the HTCondor CE.

lfield:
Can this change now be merged?

brianhlin:
See my question:

> If this should be the new provider, we should just package it but give the ability for users to choose which provider to use. How do users configure the provider that their CE is using?

Or is `condor_ce_bdii_generate_glue2.py` no longer required and all users should use the new `htcondor-ce-provider`?

lfield:
I would recommend only using the new provider. We have been running it in production for a few months now.

lfield:
That would be at CERN.

brianhlin:
If that's the case, can you replace the contents of `condor_ce_bdii_generate_glue2.py` (and rename it to `htcondor-ce-provider` if you're set on the name) so that we maintain a clear git history?

lfield:
I have related the contents of condor_ce_bdii_generate_glue2.py with htcondor-ce-provider, renamed and removed the glue1 provider.

brianhlin:
The CI tests are failing due to issues with the packaging, could you also update the packaging to reference the new provider?

Thanks,
Brian

lfield:
The CI tests have been fixed.

PR #177, opened 2017-11-21
Create required dirs on install (SOFTWARE-2806)

brianhlin:
This solution is currently incomplete as the `chown` via CMake doesn't work because of a non-existent condor user. I think this is better handled downstream via rpm/debian packaging. Do we still want to do the ownership verification via `condor_ce_startup`?

https://jira.opensciencegrid.org/browse/SOFTWARE-2806

### DIFF
--- a/rpm/htcondor-ce.spec
+++ b/rpm/htcondor-ce.spec
@@ -177,7 +177,7 @@ Conflicts: %{name}
 %setup -q
 
 %build
-%cmake -DHTCONDORCE_VERSION=%{version} -DCMAKE_INSTALL_LIBDIR=%{_libdir} -DPYTHON_SITELIB=%{python_sitelib}
+%cmake -DHTCONDORCE_VERSION=%{version} -DSTATE_INSTALL_DIR=%{_localstatedir} -DCMAKE_INSTALL_LIBDIR=%{_libdir}  -DPYTHON_SITELIB=%{python_sitelib}
 make %{?_smp_mflags}
 
 %install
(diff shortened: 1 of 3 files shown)


brianhlin:
Added permission verification for common directories across OSG and non-OSG CEs

src/condor_ce_startup
```diff
@@ -4,13 +4,19 @@ fail() {
...
     exit $RETVAL
 }
 
 # Source condor-ce environment
 [ -f /usr/share/condor-ce/condor_ce_env_bootstrap ] && . /usr/share/condor-ce/condor_ce_env_bootstrap
 
+# Verify that required dirs are owned by condor:condor (SOFTWARE-2806``)
+required_dirs="/var/run/condor-ce /var/log/condor-ce /var/log/condor-ce/user /var/lib/condor-ce /var/lib/condor-ce/spool
+/var/lib/condor-ce/execute /var/lock/condor-ce /var/lock/condor-ce/user"
+bad_dirs=$(stat --format "%n (%U:%G)" $required_dirs | awk '$2 != "(condor:condor)" {print $0}')
```
edquist: Another trick is to use `find` for file stat checks, eg:
```
bad_dirs=$(find $required_dirs -maxdepth 0 ! \( -user condor -group condor \) )
```
although the `stat ... | awk ...` solution works even when no `condor` user/group exists.

rpm/htcondor-ce.spec
```diff
@@ -190,21 +190,6 @@ rm $RPM_BUILD_ROOT%{_unitdir}/condor-ce{,-collector}.service
...
-install -m 1777 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/log/condor-ce/user
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/condor-ce
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/condor-ce/spool
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/condor-ce/spool/ceview
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/condor-ce/spool/ceview/vos
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/condor-ce/spool/ceview/metrics
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/condor-ce/execute
-install -m 0755 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lock/condor-ce
-install -m 1777 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lock/condor-ce/user
-install -m 1777 -d -p $RPM_BUILD_ROOT/%{_localstatedir}/lib/gratia/condorce_data
```
edquist: Does the build still succeed without all these `install`s in the `%install` section?  (I get that the idea is to do the dir creation in cmake instead, just wondering if it works as expected.)

brianhlin: We build the RPM in the CI, so yes: https://travis-ci.org/opensciencegrid/htcondor-ce/builds/423088342?utm_source=github_status&utm_medium=notification

edquist: Sounds good then.

edquist:
> Do we still want to do the ownership verification via `condor_ce_startup`?

Not sure what the right answer is here.  Is it possible to configure concor/htcondor-ce to use some username other than `condor`?  Because if so that will of course break this check.

brianhlin:
I think a lot more will break if they change the username of the condor/condor-ce services :)

PR #181, opened 2017-11-22
CEView daemon doesn't have permission to send keepalive to CE Master

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2820

### Symptom

Users were seeing their CE Views getting killed after 24 hours (the security session default lifetime) 

### Cause

Since the CE View starts up as root, it has access to the hostkey so it could send keepalives until the initial security session died. After it tries to get a new security session, it's running as the 'condor' user so it can't access the hostkey, can't send keepalives, and the master kills it. 

### Solution

Prefer FS auth over GSI for the master and only auth the CE View via FS. We don't fix this for the central collector because the OSG cc uses IP load balancing and certs with SANs, which don't work in HTCondor for auth (https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6135,86).

### DIFF
--- a/config/05-ce-view-defaults.conf
+++ b/config/05-ce-view-defaults.conf
@@ -39,8 +39,18 @@ SCHEDD_CRON_CEVIEW_KILL = True
 SCHEDD_CRON_CEVIEW_PREFIX = CEView
 SCHEDD_CRON_CEVIEW_RECONFIG_RERUN = True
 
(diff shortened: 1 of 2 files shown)


brianhlin:
Note that this should not need to backported to the v2 branch because of the condor 8.4 default for negotation is optional

PR #192, opened 2018-04-11
require voms-clients-cpp explicitly (SOFTWARE-3201)

edquist:
The requirement for /usr/bin/voms-proxy-info can be satisfied by
either voms-clients-cpp or voms-clients-java, the latter of which
is now problematic for osg-3.3.  So, we'll just be explicit now
about which package to bring in.


edquist:
Just a note - voms-clients-java is not currently causing any problems for us in osg-3.4, but @brianhlin suggested we go ahead and make the change in `master` also, to avoid potentially getting bitten there in the future.

PR #194, opened 2018-06-05
Build the BDII sub-package for everything but the OSG

brianhlin:
CC-IN2P3 uses BDII and grabs HTCondor-CE from the condor repos


matyasselmeci:
Did Tim make the corresponding change for the spec file in the Condor repo?

brianhlin:
We don't have the CE spec file in the condor repo

PR #201, opened 2018-08-29
Remove PhysicalCPUs

lfield:
This is the total number of physical CPUs (one physical CPU per socket) and is not the same as number of cores.


lfield:
The official document is https://www.ogf.org/documents/GFD.147.pdf
There is also our unofficial html version http://glue20.web.cern.ch/glue20

PR #206, opened 2018-08-29
Rely on WN '/bin/env' for the trace command (SOFTWARE-3387)

brianhlin:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-3387

Also part of the problem on Fermicloud VMs was the location for env that got added to the job submission existed on EL7 but not on EL6:

```
$ which env
/usr/kerberos/bin/env
```


matyasselmeci:
Should be /usr/bin/env, not /bin/env.

Fermicloud VMs have an annoying property that /usr/kerberos/bin is a symlink to /usr/bin and comes before /usr/bin in $PATH.

PR #207, opened 2018-08-30
Fix CE-ready sentinel

brianhlin:


tests/test_inside_docker.sh
```diff
@@ -35,19 +35,15 @@ function run_integration_tests {
...
     # wait until the schedd is ready before submitting a job
     for service in condor condor-ce; do
-        if [ "${OS_VERSION}" -ge 7 ]; then
-            status_check='systemctl is-active ${service}'
-        else
-            status_check='service ${service} status'
-        fi
-        timeout 30 bash -c "until (${status_check}); do sleep 0.5; done" > /dev/null 2>&1
+        timeout 30 bash -c "until (grep 'JobQueue hash' /var/log/${service}/SchedLog); do sleep 0.5; done" > /dev/null 2>&1
+        timeout 30 bash -c "until (grep 'ScheddAd' /var/log/${service}/CollectorLog); do sleep 0.5; done" > /dev/null 2>&1
```
edquist: So, as i mentioned in person, it makes me nervous that the presence of strings `JobQueue hash` or `ScheddAd` in log files can serve as a test condition; but since this is always happening on a fresh VM i guess it's ok.

Style-wise, if you just want to test the presence of a string with `grep`, rather than actually printing all the lines that contain it, it's better to do `grep -q`.  You can also then forget about the redirects to `/dev/null`.  (Presumably it doesn't matter all that much here, but if you have large log files, another advantage of `grep -q` is that it will return immediately after the first match, rather than continuing to process all of the input file(s).)

PR #208, opened 2018-09-07
Revert the default condor_q behavior to query all users' jobs

brianhlin:
The new condor_q behavior requires an authZ check and that can cause
failures when running condor_ce_q. We don't care who can query the CE
schedd so this isn't a security issue.


matyasselmeci:
Hi @sthapa, do you have time to review this soon? I'd like to include it for next week's OSG release.

matyasselmeci:
Looks like a simple enough fix. Merged.

PR #210, opened 2018-11-27
Map certs with VOMS attr before local daemons (SOFTWARE-3489)

brianhlin:


config/condor_mapfile.osg
```diff
@@ -1,3 +1,4 @@
+GSI ".*,/[A-Za-z0-9]*/Role=[A-Za-z0-9]*/Capability=NULL.*" GSS_ASSIST_GRIDMAP
```
bbockelm: I'm not sure about the `.*` at the end -- can we have `$` instead?

bbockelm: Ah, also, this might not work with nested groups (`/cms/uscms`).

brianhlin: I think that would possibly break FIFE jobs. IIRC, they have a ton of VOMS FQANs

brianhlin: Good point about the nested groups. How deep can they go?

bbockelm: Pretty deep.  I would just allow `/` in the character specification.

brianhlin: I've fixed the match expression to support nested groups and limited the matches for subsequent VOMS FQANs. Are there any other characters that are supported in VO names, group names, roles, and capabilities that we need to add?

config/condor_mapfile.osg
```diff
@@ -1,3 +1,4 @@
+GSI ".*,/[/A-Za-z0-9\.]*/Role=[A-Za-z0-9]*/Capability=NULL[,/A-Za-z0-9\.=]*$" GSS_ASSIST_GRIDMAP
```
bbockelm:
```suggestion
GSI ".*,/[/A-Za-z0-9\.]*/Role=[A-Za-z0-9\.]*/Capability=NULL$" GSS_ASSIST_GRIDMAP
```

Two small tweaks:
1. I think `.` is valid for Roles.
2. Nothing is permitted past the capability.

brianhlin: > Nothing is permitted past the capability.

Except additional VOMS FQANs, which come in comma separated

brianhlin: Oh duh, it just matches the last VOMS FQAN and any preceding ones are caught in the `.*`.

PR #214, opened 2019-01-08
production hardening

lfield:
Some improvements from running in production

- Adding a timeout on CE queries
- Adding a try/catch

### DIFF
--- a/contrib/bdii/htcondor-ce-provider
+++ b/contrib/bdii/htcondor-ce-provider
@@ -5,8 +5,9 @@ import sys
 from datetime import datetime
 import subprocess
 import collections
-
+import time
 import htcondor
+import signal
 
 service_ldif = """dn: {bind_dn}
 GLUE2ServiceID: {central_manager}
@@ -100,6 +101,8 @@ GLUE2PolicyRule: vo:{vo}
 GLUE2MappingPolicyShareForeignKey: {shareid}
 """
 
+def handler(signum, frame):
+    raise Exception("TimeoutError")
 
 def main():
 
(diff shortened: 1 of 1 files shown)


contrib/bdii/htcondor-ce-provider
```diff
@@ -100,6 +101,8 @@ GLUE2PolicyRule: vo:{vo}
 GLUE2MappingPolicyShareForeignKey: {shareid}
 """
 
+def handler(signum, frame):
+    raise Exception("TimeoutError")
```
brianhlin: Could you raise `RuntimeError` or better yet, create a `TimeoutException` class?

contrib/bdii/htcondor-ce-provider
```diff
@@ -128,21 +131,24 @@ def main():
...
+                    topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']] = {}
+                if ad['OpSysMajorVer'] not in topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']]:
+                    topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']][ad['OpSysMajorVer']] = {}
+                if ad['DetectedCpus'] not in topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']][ad['OpSysMajorVer']]:
+                    topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']][ad['OpSysMajorVer']][ad['DetectedCpus']] = {}
+                if ad['DetectedMemory'] not in topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']][ad['OpSysMajorVer']][ad['DetectedCpus']]:
+                    topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']][ad['OpSysMajorVer']][ad['DetectedCpus']][ad['DetectedMemory']] = 0
+                topologies[ad['Arch']][ad['OpSys']][ad['OpSysName']][ad['OpSysMajorVer']][ad['DetectedCpus']][ad['DetectedMemory']] += 1
+        except KeyError:
+            pass
```
brianhlin: I think this entire block can be simplified with a series of default dicts (I can submit a PR). Do you know which key actually had issues? Catching all `KeyErrors` here seems like a hammer to solve a potential logic problem

contrib/bdii/htcondor-ce-provider
```diff
@@ -208,6 +216,11 @@ def main():
         except (KeyError, RuntimeError):
             state = 'critical'
             state_info = 'Authorization ping failed'
+        except Exception, e:
```
brianhlin: Let's catch either `RuntimeError` (and verify the TimeoutError message, raising if the message doesn't match) or `TimeoutException` depending on what you do above.

contrib/bdii/htcondor-ce-provider
```diff
@@ -221,14 +234,22 @@ def main():
...
+        try:
+            for job in query:
+                if not job.get("JobStatus") or not job.get("x509userproxyvoname"):
+                    continue
+                total_vo_jobs[job['x509userproxyvoname']] += 1
+                if job['JobStatus'] == 1:
+                    idle_vo_jobs[job['x509userproxyvoname']] += 1
+                elif job['JobStatus'] == 2:
+                    running_vo_jobs[job['x509userproxyvoname']] += 1
+        except Exception, e:
```
brianhlin: Same thing about catching `RuntimeError` or `TimeoutException`

contrib/bdii/htcondor-ce-provider
```diff
@@ -199,6 +205,8 @@ def main():
             sys.stderr.write("Unable to locate CE schedd on %s" % ce_host)
             continue
 
+        signal.signal(signal.SIGALRM, handler)
+        signal.alarm(10)
```
brianhlin: Can we make the timeout window based off an HTCondor configuration value with a default of 10s? If you don't know how to do this, I can submit a PR

contrib/bdii/htcondor-ce-provider
```diff
@@ -208,6 +216,11 @@ def main():
         except (KeyError, RuntimeError):
             state = 'critical'
             state_info = 'Authorization ping failed'
+        except Exception, e:
+            sys.stderr.write("Ping to CE schedd on %s timed out.\n" % ce_host )
```
brianhlin:
```suggestion
            sys.stderr.write("Ping to CE schedd on %s timed out after 10s.\n" % ce_host )
```

contrib/bdii/htcondor-ce-provider
```diff
@@ -221,14 +234,22 @@ def main():
...
+            for job in query:
+                if not job.get("JobStatus") or not job.get("x509userproxyvoname"):
+                    continue
+                total_vo_jobs[job['x509userproxyvoname']] += 1
+                if job['JobStatus'] == 1:
+                    idle_vo_jobs[job['x509userproxyvoname']] += 1
+                elif job['JobStatus'] == 2:
+                    running_vo_jobs[job['x509userproxyvoname']] += 1
+        except Exception, e:
+            sys.stderr.write("CE schedd on %s timed out.\n" % ce_host )
```
brianhlin:
```suggestion
            sys.stderr.write("CE schedd on %s timed out after 10s.\n" % ce_host )
```

contrib/bdii/htcondor-ce-provider
brianhlin: Actually I think there's better exception handling that we can do here, too. This is fine as is but I'll submit another PR to address it

contrib/bdii/htcondor-ce-provider
lfield: Add will add a TimeoutError Exception class

contrib/bdii/htcondor-ce-provider
lfield: The error was with OpSysMajorVer but any attribute could be missing from a malformed ad.

contrib/bdii/htcondor-ce-provider
lfield: Going to catch the TimeoutError Exception.

contrib/bdii/htcondor-ce-provider
lfield: There is no timeout on this query. If the sched was slow this would have been discovered with the earlier ping.

contrib/bdii/htcondor-ce-provider
lfield: Doing this change with the new time_out variable

contrib/bdii/htcondor-ce-provider
lfield: Will add this option

lfield:
I have merged your PR and address all the comments.

contrib/bdii/htcondor-ce-provider
```diff
@@ -208,6 +221,11 @@ def main():
         except (KeyError, RuntimeError):
             state = 'critical'
             state_info = 'Authorization ping failed'
+        except TimeoutError, e:
+            sys.stderr.write("Ping to CE schedd on %s timed out after %i s.\n" % ce_host, timeout )
```
brianhlin:
```suggestion
            sys.stderr.write("Ping to CE schedd on %s timed out after %i s.\n" % ce_host, time_out )
```

contrib/bdii/htcondor-ce-provider
```diff
@@ -119,6 +125,11 @@ def main():
         sys.stderr.write("Error: GLUE2DomainID: not set\n")
         sys.exit(1)
 
+    # Get the timeout value
+    time_out = htcondor.param.get('CE_QUERY_TIMEOUT')
```
brianhlin: Can we change this to `BDII_CE_QUERY_TIMEOUT` and add `BDII_CE_QUERY_TIMEOUT = 10` to each of the config files in the `bdii` dir?

brianhlin: Could you expand on that? I was under the impression that GLUE was basically the schema and BDII is the system by which GLUE information is reported. So maybe BDII is a misnomer for this entire package but I'd rather have the naming be consistent for the user.

lfield: The BDII is an information collection/aggregation service. GLUE is an information model and serialization into different data formats. This code is an information provider that queries the HTCondor batch system and returns information according to the GLUE2 information model in the LDIF.  I hope that is clear. I hope that is clear so yes the directory should be glue rather than bdii. For example you could imaging this provider being modified in the future to output the information in the JSON format to be compatible with CRIC.

brianhlin: I would think that if we were to support the JSON/CRIC case, we'd call it the CRIC provider because that better describes the schema and the consumer of the information. If this provider starts doing double-duty, we should probably rename it all to `htcondor-ce-info-provider` or something along those lines.

Does anyone use GLUE2 outside of BDII?

brianhlin:
I'm not going to hold up the PR any longer for a minor variable naming issue.

contrib/bdii/htcondor-ce-provider
lfield: Yes, if the information model was specific to CRIC then yes it would be a different provider. GLUE2 is used outside of the BDII. There are official XML and JSON rendering documents.

brianhlin: Sorry, I meant that we'd call it the "CRIC provider" instead of the "JSON provider". I think the same logic works here, especially since BDII is the only consumer of GLUE2 information: although the CE query is for constructing GLUE2-compliant data, it is ultimately for BDII service. I would think the end-user would be more concerned/aware of the service over the data format itself so it'd be friendlier to name the package, config, etc. after the service.

PR #217, opened 2019-02-25
Use non-IU specific URL

brianhlin:


config/01-ce-collector.conf
```diff
@@ -8,7 +8,7 @@
 #
 ###############################################################################
 
-MYOSG_HOST = myosg.grid.iu.edu
+MYOSG_HOST = my.opensciencegrid.org
 MYOSG_URL = http://$(MYOSG_HOST)/rgsummary/xml?summary_attrs_showservice=on&all_resources=on&gridtype=on&gridtype_1=on&service=on&service_sel%5B%5D=1&active=on&active_value=1&disable=on&disable_value=0
```
matyasselmeci: As long as you're changing the host, let's fix the query, too.
```suggestion
MYOSG_URL = http://$(MYOSG_HOST)/rgsummary/xml?gridtype=on&gridtype_1=on&service=on&service_sel%5B%5D=1&active=on&active_value=1&disable=on&disable_value=0
```

PR #220, opened 2019-03-26
Update job router configuration to use new HTCondor multi-line syntax (SOFTWARE-3637)

brianhlin:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-3639


brianhlin:
@timtheisen we could use your HTCondor expertise on this review

PR #223, opened 2019-04-12
Adding default memory to queue information

djw8605:
This failed for sites which did not set memory in the
resource catalog.

### DIFF
--- a/src/htcondorce/web_utils.py
+++ b/src/htcondorce/web_utils.py
@@ -127,7 +127,7 @@ def generate_queue_ad(resource_catalog, ce):
                          'entry': entry.get('Name', ''),
                          'name': queue,
                          'status': 'Production',
-                         'memory': int(entry.get('Memory')),
+                         'memory': int(entry.get('Memory', 0)),
                          'cpus': int(entry.get('CPUs', 1)),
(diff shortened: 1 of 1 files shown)


brianhlin:
Shouldn't they be setting the memory in the resource catalog, though?

djw8605:
Yes, but it shouldn't break the webpage when they do not.

brianhlin:
Sure but I we should be directing the user to fix the root cause of the problem

brianhlin:
@djw8605 is there an easy way to display a warning/error banner with CherryPy?

djw8605:
This is put in a JSON that is parsed by a machine.  No human should be reading this output, and no human would see this warning.

brianhlin:
Sorry, I'm confused. You mentioned above that without a default a webpage is breaking, which page was that?

djw8605:
The JSON that AGIS consumes.  This site: http://collector.opensciencegrid.org/json/agis-compat

djw8605:
Good idea on the failed-ce's.  We now need a test to make sure failed-ce's is empty.  We have check_mk testing the agis-compat URL, just another check_mk test.

PR #224, opened 2019-05-23
Move APEL configuration into the CE config

brianhlin:
Hey @sjones-hep-ph-liv-ac-uk, here's a first attempt at pushing the APEL config from a separate file into the CE configuration itself.


contrib/apelscripts/condor_batch.sh
```diff
@@ -8,21 +8,12 @@ output=batch-$(date -u --date='yesterday' +%Y%m%d )-$(hostname -s)
...
-trap '' ERR
-
-[[ -z "$SCALING_ATTR" ]] && SCALING_ATTR=MachineAttrRalScaling0
-[[ -z "$BATCH_HOST" ]] && BATCH_HOST=`hostname`
-[[ -z "$OUTPUT_LOCATION" ]] && OUTPUT_LOCATION=/var/lib/condor-ce/apel/
-
-OUTPUT_FILE=$OUTPUT_LOCATION/$output
+HISTORY_SUFFIX='-format "\n" EMPTY'
+SCALING_ATTR=$(condor_ce_config_val APEL_SCALING_ATTR)
+[ $? -eq 0] && HISTORY_SUFFIX="-format "%v|" ${SCALING_ATTR} ${HISTORY_SUFFIX}"
```
matyasselmeci:
```suggestion
[ $? -eq 0 ] && HISTORY_SUFFIX="-format "%v|" ${SCALING_ATTR} ${HISTORY_SUFFIX}"
```

contrib/apelscripts/condor_batch.sh
```diff
@@ -8,21 +8,12 @@ output=batch-$(date -u --date='yesterday' +%Y%m%d )-$(hostname -s)
...
 
 TZ=GMT condor_history -constraint "$CONSTR" \
-    -format "%s_${BATCH_HOST}|" ClusterId \
+    -format "%s_$(condor_ce_config_val APEL_BATCH_HOST)|" ClusterId \
```
matyasselmeci: What happens if this or APEL_OUTPUT_DIR do not exist or have bogus values?

contrib/apelscripts/condor_batch.sh
```diff
@@ -32,6 +23,5 @@ TZ=GMT condor_history -constraint "$CONSTR" \
     -format "%d|" ResidentSetSize_RAW \
     -format "%d|" ImageSize_RAW \
     -format "%d|" RequestCpus \
-    -format "%v|" ${SCALING_ATTR} \
-    -format "\n" EMPTY  | sed -e "s/undefined|$/1.0|/" > $OUTPUT_FILE
+    ${HISTORY_SUFFIX} > $(condor_ce_config_val APEL_OUTPUT_DIR)/$output
```
matyasselmeci:
```suggestion
    "${HISTORY_SUFFIX}" > "$(condor_ce_config_val APEL_OUTPUT_DIR)/$output"
```
but you should probably get APEL_OUTPUT_DIR separately and deal with it being empty.

contrib/apelscripts/condor_batch.sh
brianhlin: We provide default config values in `/usr/share/condor-ce/config.d/` so I'm not as worried about the values not existing

contrib/apelscripts/condor_batch.sh
```diff
@@ -32,6 +23,5 @@ TZ=GMT condor_history -constraint "$CONSTR" \
     -format "%d|" ResidentSetSize_RAW \
     -format "%d|" ImageSize_RAW \
     -format "%d|" RequestCpus \
-    -format "%v|" ${SCALING_ATTR} \
-    -format "\n" EMPTY  | sed -e "s/undefined|$/1.0|/" > $OUTPUT_FILE
+    "${HISTORY_SUFFIX}" > "$(condor_ce_config_val APEL_OUTPUT_DIR)/$output"
```
edquist: This doesn't look right to me.  Why is `"${HISTORY_SUFFIX}"` quoted?

Hint: I think maybe you want to construct it as an array instead of a single string.

contrib/apelscripts/condor_batch.sh
matyasselmeci: What is the behavior when the condor ce config is unparseable, for example due to a syntax error?

contrib/apelscripts/condor_batch.sh
```diff
@@ -23,5 +30,4 @@ TZ=GMT condor_history -constraint "$CONSTR" \
     -format "%d|" ResidentSetSize_RAW \
     -format "%d|" ImageSize_RAW \
     -format "%d|" RequestCpus \
-    "${HISTORY_SUFFIX}" > "$(condor_ce_config_val APEL_OUTPUT_DIR)/$output"
-
+    "${HISTORY_SUFFIX}" > $OUTPUT_FILE
```
edquist: Same comment as before - it seems to me that `"${HISTORY_SUFFIX}"` should not be quoted as a single string.

If you think you need quoting for some reason (to escape whitespace or glob characters?), then consider constructing `HISTORY_SUFFIX` as an array and then quoting it as `"${HISTORY_SUFFIX[@]}"`

edquist: (@matyasselmeci also suggests a less confusing variable name might also be `HISTORY_EXTRA_ARGS`)

contrib/apelscripts/condor_batch.sh
```diff
@@ -3,26 +3,24 @@
...
-[[ -z "$SCALING_ATTR" ]] && SCALING_ATTR=MachineAttrRalScaling0
-[[ -z "$BATCH_HOST" ]] && BATCH_HOST=`hostname`
-[[ -z "$OUTPUT_LOCATION" ]] && OUTPUT_LOCATION=/var/lib/condor-ce/apel/
+# Build the filter for the history command
+CONSTR="EnteredCurrentStatus >= $yesterday && EnteredCurrentStatus < $today && RemoteWallclockTime !=0"
 
-OUTPUT_FILE=$OUTPUT_LOCATION/$output
+HISTORY_SUFFIX='-format "\n" EMPTY'
+SCALING_ATTR=$(condor_ce_config_val APEL_SCALING_ATTR)
+[ $? -eq 0 ] && HISTORY_SUFFIX="-format "%v|" ${SCALING_ATTR} ${HISTORY_SUFFIX}"
```
edquist: Note that the `|` here is completely unquoted and will be interpreted as a pipeline to be executed in this line

```suggestion
[ $? -eq 0 ] && HISTORY_SUFFIX="-format \"%v|\" ${SCALING_ATTR} ${HISTORY_SUFFIX}"
```

edquist: (Credit @matyasselmeci for catching this one)

contrib/apelscripts/condor_batch.sh
```diff
@@ -1,28 +1,30 @@
 #!/bin/bash
 
+safe_config_val () {
+    condor_ce_config_val $1 || echo "Failed to retrieve CE configuration value '$1'" && exit 1
```
edquist: I dunna think this line does what you think it does.

```
$ echo Success || echo Badness && echo I give up
```

contrib/apelscripts/condor_batch.sh
```diff
@@ -1,28 +1,32 @@
...
 
-[[ -z "$SCALING_ATTR" ]] && SCALING_ATTR=MachineAttrRalScaling0
-[[ -z "$BATCH_HOST" ]] && BATCH_HOST=`hostname`
-[[ -z "$OUTPUT_LOCATION" ]] && OUTPUT_LOCATION=/var/lib/condor-ce/apel/
+# Build the filter for the history command
+CONSTR="EnteredCurrentStatus >= $yesterday && EnteredCurrentStatus < $today && RemoteWallclockTime !=0"
 
-OUTPUT_FILE=$OUTPUT_LOCATION/$output
+HISTORY_EXTRA_ARGS='-format "\n" EMPTY'
+SCALING_ATTR=$(safe_config_val APEL_SCALING_ATTR)
```
edquist: Ah, but it's not so easy because `safe_config_val` is run in a subshell, and the `exit 1` won't be propagated...

I can put together a suggestion...

matyasselmeci: Wanna bet?

matyasselmeci: Never mind. misunderstood the pr again.

brianhlin:
@edquist are we good to go here?

edquist:
The only thing left i would ask is if anyone has actually tried out the code to see if it works, but otherwise i don't see any more obvious problems.

brianhlin:
Stephen tested out the changes and said they were all good!

PR #225, opened 2019-06-13
 Update the central collector via TCP (SOFTWARE-3668) 

brianhlin:
https://opensciencegrid.atlassian.net/browse/SOFTWARE-3668

### DIFF
--- a/config/05-ce-view-defaults.conf
+++ b/config/05-ce-view-defaults.conf
@@ -56,8 +56,7 @@ endif
 # with SIGKILL signal (9) after waiting 10 seconds
 SHUTDOWN_FAST_TIMEOUT = 10
 
-# If available, run CE View as a DC daemon.
-if version >= 8.3.5
-  DC_DAEMON_LIST = +CEVIEW
-endif
+# Run the CE View as a DC daemon.
+DC_DAEMON_LIST = +CEVIEW
+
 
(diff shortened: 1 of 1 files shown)


config/05-ce-view-defaults.conf
```diff
@@ -56,8 +56,12 @@ endif
...
-  DC_DAEMON_LIST = +CEVIEW
-endif
+# Run the CE View as a DC daemon.
+DC_DAEMON_LIST = +CEVIEW
 
+# Improve consistency of central collector updates
+# Prior to 8.9.2, these updates blocked
+# https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6249
+if version >= 8.9.2
+  UPDATE_VIEW_COLLECTOR_WITH_TCP = True
```
matyasselmeci: didn't we do this once already?

brianhlin: We update the site CE collector via TCP (https://github.com/opensciencegrid/htcondor-ce/blob/6b9f52c46ac3a06a3f019d896515c606ef53a319/config/01-ce-info-services-defaults.conf#L2) but we don't currently do this for the central collector (https://github.com/opensciencegrid/htcondor-ce/search?q=TCP&unscoped_q=TCP)

PR #229, opened 2019-07-31
Deploy built (S)RPMs to a server at CHTC

matyasselmeci:


brianhlin:
It looks like the deploy tests ran and they built the RPM. Why bother if we're not actually deploying the binray RPMs?

matyasselmeci:
Good question. Let me futz around and see if I can make travis not do that...

.travis.yml
```diff
@@ -1,24 +1,54 @@
...
+    - <<: *run_tests
+      env:
+        - OS_TYPE=centos OS_VERSION=7 BUILD_ENV=osg
 
+    - &deploy_rpm
+      # Template; subsequent uses modify 'env'
+      env:
+        - OS_TYPE=centos OS_VERSION=6 BUILD_ENV=uw_build
+      on:
+        repo: htcondor/htcondor-ce
```
brianhlin: Looks like the deploy tasks still ran. Can't we combine `on` directives like so?

```suggestion
        repo: htcondor/htcondor-ce
        tags: true
        branch: master
```

matyasselmeci: Last time I tried it, `tags: true` and `branch: master` didn't go together: I tried making a tag and it didn't do the deploy because 'v3.2.2.5matyas is not master'. I can add `tags: true` and make it only deploy on tags.

matyasselmeci:
on:tags:true now shows the build as 'canceled' in travis and errored in github.

brianhlin:
Well that's dumb >.>.

Let's do what we do in the OSG doc deployments, i.e. handle it in the script(s) itself: https://github.com/opensciencegrid/doc-ci-scripts/blob/master/deploy

matyasselmeci:
Good grief, is that really what we do? :man_facepalming: 
I'll give that a try.

.travis.yml
brianhlin: So as it stands now, we'll get RPMs on every commit to the repo? Is that what @timtheisen wants?

matyasselmeci: ?
No, that's not how it works.

matyasselmeci: At least, RPMs won't be created on PRs until they're merged. If you want me to only do it on tags, I can do that too.

matyasselmeci:
OK, I think this is now correctly uploading on tags and correctly not uploading on not-tags.

matyasselmeci:
I swear the first one already works for GCT.

brianhlin:
It looks like builds (10-15 min each D:) occur on every commit, they just skip the deployment part unless it's a tag: https://travis-ci.org/gridcf/gct/builds/565402469

matyasselmeci:
\*sigh\*

PR #233, opened 2019-08-19
Try to determine which routes are misconfigured

LincolnBryant:
Our HTCondor-CE has a lot of complex routes which can be difficult to debug. This PR is kind of conceptual to try to help a site admin to determine which route is bad.

It's been a while since I hacked this together, but as I recall it will try to validate the routes one-by-one, and then subtract the set of known good routes from all routes and produce a set of possibly bad ones. 

This produces some additonal output to try to narrow down the bad config, e.g.:
```
[root@iut2-gk ~]#  /usr/share/condor-ce/condor_ce_env_bootstrap && ./verify.py 
Could not read JOB_ROUTER_ENTRIES in the HTCondor CE configuration. Please verify syntax correctness
ERROR: The following appear to be malformed: DefaultAtlas, OpportunisticVO
```

It does produce some false positives (i.e., in this case Default Atlas was fine, but Opportunistic VO had some missing parens), but it narrowed the search space from 15 routes to 2. 

Just some ideas.

### DIFF
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -14,6 +14,26 @@ def error(msg):
 def warn(msg):
     print "WARNING: " + msg
 
+def condorpp(param):
+    syntax = [';','[',']']
+    out = "".join(param.split()) # strip all whitespace
+    for elem in syntax: 
+      out = out.replace(elem, elem+'\n') # re-print 
+    return out
+
+def malformedQueues(parsedAds, unparsedAds):
+
+    allqueues = []
+    for line in condorpp(unparsedAds).split('\n'):
+        if 'name' in line:
+            allqueues.append(line.split('=')[1].split(';')[0].strip('\"'))
+
+    goodqueues = []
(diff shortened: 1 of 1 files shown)


brianhlin:
Hi @LincolnBryant I know this is really late but do you have the DefaultAtlas and OpportunisticVO routes that triggered the error?

LincolnBryant:
In that particular case, I was missing a paren or square bracket as I recall. So the script just evaluated things until it hit something it didn't like. It's definitely not all-encompassing but it did help narrow down the problem area.

brianhlin:
@LincolnBryant just so I'm not leaving you in the dark, I've got a WIP branch that refactors this a bit so that the message points you to the first route with an error it finds

brianhlin:
And a JIRA ticket to track this work since I didn't get the chance to finish my WIP branch: https://opensciencegrid.atlassian.net/browse/SOFTWARE-3896

PR #236, opened 2019-09-04
Reorganize central collector configuration (SOFTWARE-3289)

brianhlin:
- Move generated config to /usr/share/condor-ce/config.d
- Create /etc/condor-ce/02-ce-collector.conf for the admin to manage
  advertising authz
- Admins shouldn't /etc/condor-ce/config.d/04-ce-collector.conf, which
  allows all CEs to advertise to the central collector via GSI.
- Other misc config fixups


config/02-ce-collector-auth.conf
```diff
@@ -0,0 +1,26 @@
...
+
+# To BAN a host, add a line of the form:
+#
+#   DENY_ADVERTISE_SCHEDD = $(DENY_ADVERTISE_SCHEDD), ce.example.com@users.htcondor.org
+#
+# You can have multiple comma-separated hosts in the DENY listing.
+
+# To ADD a custom host, add a line of the form:
+#
+#   COLLECTOR.ALLOW_ADVERTISE_SCHEDD = $(COLLECTOR.ALLOW_ADVERTISE_SCHEDD), ce.example.com@users.htcondor.org/ce.example.com
```
matyasselmeci: Why COLLECTOR.ALLOW_ADVERTISE_SCHEDD instead of ALLOW_ADVERTISE_SCHEDD? Conversely, above, why DENY_ADVERTISE_SCHEDD instead of COLLECTOR.DENY_ADVERTISE_SCHEDD?

matyasselmeci:
Why are we shipping a 'generated' conf at all instead of generating it on first run or something?

config/02-ce-collector-auth.conf
brianhlin: @bbockelm originally wrote this config but my interpretation is that we only want to allow schedds to advertise to the collector but when we want a host banned, it should be banned completely.

brianhlin:
I think that'd be because it was annoying to do via SysVinit but would be easier to do for just Systemd. But we'd have to ask @bbockelm to see if he can remember the details

src/condor_ce_config_generator
```diff
@@ -210,28 +210,8 @@ CONFIG_TEMPLATE = """\
...
-#
-# You can have multiple comma-separated hosts in the ALLOW listing.
-
-# The generated ALLOW_ADVERTISE_SCHEDD should be of the form:
-#
-#   COLLECTOR.ALLOW_ADVERTISE_SCHEDD = foo.example.com@users.opensciencegrid.org/foo.example.com, bar.example.com@users.opensciencegrid.org/bar.example.com
-#
-# There will be one entry per enabled service found in MyOSG.  This file is updated
-# atomically; MyOSG outages will not result in partial or incorrect updates.
+# There will be one entry per active, registered CE in OSG Topology.  This file is updated
```
matyasselmeci: Is this comment going to be inside the generated file? If so, I think the first sentence is the only one that is relevant to the admin; the rest are implementation details.

brianhlin: I agree, the original author was overly verbose here :P

matyasselmeci:
> I think that'd be because it was annoying to do via SysVinit but would be easier to do for just Systemd. But we'd have to ask @bbockelm to see if he can remember the details

I'd like to do that. Files that are supposed to be generated shouldn't get shipped. Are these changes for OSG 3.5 only, or will they be backported to 3.4 as well?

brianhlin:
3.5 only and fine but let's do it in a separate PR

matyasselmeci:
Sure.

config/02-ce-collector-auth.conf
```diff
@@ -30,13 +24,3 @@
 # The generated ALLOW_ADVERTISE_SCHEDD should be of the form:
```
matyasselmeci:
```suggestion
# The COLLECTOR.ALLOW_ADVERTISE_SCHEDD lines should be of the form:
```

src/condor_ce_config_generator
```diff
@@ -209,29 +209,7 @@ CONFIG_TEMPLATE = """\
...
-#
-# You can have multiple comma-separated hosts in the ALLOW listing.
-
-# The generated ALLOW_ADVERTISE_SCHEDD should be of the form:
-#
-#   COLLECTOR.ALLOW_ADVERTISE_SCHEDD = foo.example.com@users.opensciencegrid.org/foo.example.com, bar.example.com@users.opensciencegrid.org/bar.example.com
-#
-# There will be one entry per enabled service found in MyOSG.  This file is updated
-# atomically; MyOSG outages will not result in partial or incorrect updates.
+# There will be one entry per active, registered CE in OSG Topology.
```
matyasselmeci:
```suggestion
# There is one entry per active, registered CE in OSG Topology.
```

PR #238, opened 2019-09-06
Automatically enable CEView; consolidate configs (SOFTWARE-3289)

brianhlin:
@djw8605 how does automatically enabling the CEView sound to you?


djw8605:
It's fine with me.  It opens another port on the host, I think by default, port 80.

brianhlin:
Yup, though when I said automatic, I meant "automatic upon htcondor-ce-view package installation". Since that package is optional for CE installations, this won't enable it/open ports for all installs.

PR #242, opened 2019-09-10
Disable job retries (SOFTWARE-3407)

brianhlin:
This is the job retry portion of #240 to discuss this feature separately. From https://github.com/htcondor/htcondor-ce/pull/240/files#r322860688:

> @JaimeFrey I was hoping for your opinion on this periodic remove expression for disabling job retries on the CE. Two things that I can think of off the top of my head:
> 
> 1. We should only remove CE jobs (i.e. not routed grid universe jobs) based on number of job retries
> 1. We may want to avoid removing held jobs with `NumJobStarts > 1` so if there are helpful hold messages we don't remove them from under the admin immediately


brianhlin:
@jdost321 and I tested this with the ITB factory and our ITB CE succesfully.

Are there any cases where `NumJobStarts` gets updated before `JobStatus`?

brianhlin:
The job removals did end up spiking the number of held jobs on the factory side but @jdost321 suggested that they would be released and resubmitted with a new GridJobID and that this was fine for factory ops.

JaimeFrey:
The duplicate clauses for RoutedtoJobId and $(ENABLE_JOB_RETRIES) can be consolidated.
I'd like to discourage the use of the magic attribute CurrentTime; use time() instead.
Otherwise, this looks good.

PR #247, opened 2019-09-17
Added IOError exception handling

lfield:
Added IOError exception handling, generated when querying the CE Collector

### DIFF
--- a/contrib/bdii/htcondor-ce-provider
+++ b/contrib/bdii/htcondor-ce-provider
@@ -210,7 +210,6 @@ def main():
             sys.stderr.write("Unable to locate CE schedd on %s" % ce_host)
             continue
         except EnvironmentError:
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Might be a good idea to a) catch EnvironmentError instead of IOError (EnvironmentError is the parent class of IOError and OSError) and b) print the exception message (for RuntimeError above as well).

contrib/bdii/htcondor-ce-provider
```diff
@@ -209,7 +209,11 @@ def main():
         except RuntimeError:
             sys.stderr.write("Unable to locate CE schedd on %s" % ce_host)
             continue
-
+        except EnvironmentError:
+            sys.stderr.write("Unable to locate CE schedd on %s" % ce_host)
```
matyasselmeci: This error message looks like it got duplicated
```suggestion
```

PR #250, opened 2019-09-30
Add Schedd cron for APEL record generation

brianhlin:


brianhlin:
@sjones-hep-ph-liv-ac-uk note that this adds a schedd cron for the APEL record generation and upload so you don't need your own cron job.

I've also provided a default scaling attribute name but you can continue to use `RalScaling`

PR #253, opened 2019-10-02
Document how to use grid certificates for SSL/token auth

brianhlin:
Can't use 'if $(CE.USE_GRID_CERTS)' since the if statement is
evaluated at parse time and the admin configuration is parsed after
the provided default configs.

Looking at this config again, though, I think maybe we should set the following regardless:

```
AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/hostcert.pem
AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/hostkey.pem
```

Because I'm not sure I see much benefit in the change in behavior that the default values would incur:

```
AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/condor/hostcert.pem
AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/condor/hostkey.pem
```

### DIFF
--- a/config/01-ce-auth.conf
+++ b/config/01-ce-auth.conf
@@ -7,9 +7,14 @@
 #
 ###############################################################################
 
-# Uncomment the following line if your host certificate was issued by an
+# Uncomment the following lines if your host certificate was issued by an
 # IGTF-accredited CA, such as the InCommon IGTF CA or DigiCert Grid CA-1:
 # https://dl.igtf.net/distribution/igtf/current/accredited/accredited.in
 # If you are using a Let's Encrypt certificate, leave this line commented.
 #
-# CE.USE_GRID_CERTS = True
+# AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/hostcert.pem
+# AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/hostkey.pem
+# AUTH_SSL_SERVER_CADIR = /etc/grid-security/certificates
+# AUTH_SSL_CLIENT_CADIR = /etc/grid-security/certificates
+# AUTH_SSL_SERVER_CAFILE =
+# AUTH_SSL_CLIENT_CAFILE =
--- a/docs/installation/htcondor-ce.md
+++ b/docs/installation/htcondor-ce.md
@@ -159,6 +159,22 @@ as the following HTCondor-CE configuration:
 
             globus_mapping /usr/lib64/libgsi_pep_callout.so argus_pep_callout
 
+#### (Optional) Use grid certificates for SSL authentication ####
+
+By default, HTCondor-CE uses system certificates for authenticating SSL connections and expects a host certificate and
+key in `/etc/grid-security/condor/hostcert.pem` and `/etc/grid-security/condor/hostkey.pem`, respectively.
+If you or your clients are using [IGTF-accredited certificates](https://dl.igtf.net/distribution/igtf/current/accredited/accredited.in),
+set the following configuration in `/etc/condor-ce/config.d/01-ce-auth.conf`:
+
+```
+AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/hostcert.pem
+AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/hostkey.pem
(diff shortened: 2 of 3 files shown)


config/01-ce-auth.conf
```diff
@@ -7,9 +7,14 @@
...
+# Uncomment the following lines if your host certificate was issued by an
 # IGTF-accredited CA, such as the InCommon IGTF CA or DigiCert Grid CA-1:
 # https://dl.igtf.net/distribution/igtf/current/accredited/accredited.in
 # If you are using a Let's Encrypt certificate, leave this line commented.
 #
-# CE.USE_GRID_CERTS = True
+# AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/hostcert.pem
+# AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/hostkey.pem
+# AUTH_SSL_SERVER_CADIR = /etc/grid-security/certificates
+# AUTH_SSL_CLIENT_CADIR = /etc/grid-security/certificates
```
matyasselmeci: Should the client CA dir really depend on what kind of certificate the server has?

brianhlin: Yes because the clients in this case are not clients to the server but rather when the server is acting as a client

docs/installation/htcondor-ce.md
```diff
@@ -159,6 +159,22 @@ as the following HTCondor-CE configuration:
 
             globus_mapping /usr/lib64/libgsi_pep_callout.so argus_pep_callout
 
+#### (Optional) Use grid certificates for SSL authentication ####
+
+By default, HTCondor-CE uses system certificates for authenticating SSL connections and expects a host certificate and
+key in `/etc/grid-security/condor/hostcert.pem` and `/etc/grid-security/condor/hostkey.pem`, respectively.
+If you or your clients are using [IGTF-accredited certificates](https://dl.igtf.net/distribution/igtf/current/accredited/accredited.in),
```
bbockelm: IGTF certificates aren't really the deciding factor, are they?  It's more whether they use the system trust roots or the traditional grid location for trust roots.  How about:

> By default, HTCondor-CE uses the default system locations to locate certificate authorities (CAs) and host certificate when authenticating SSL connections.  If you would prefer to to use the traditional "grid location" for CAs and the host certificate, set the following configuration in `/etc/condor-ce/config.d/01-ce-auth.conf`:

?

bbockelm:
I'm a bit ambivalent on using the "grid location" for the host certificate versus the "system location".

The "system location" doesn't appear to be strongly used even by base packages (e.g., Let's Encrypt doesn't write there by default).  It introduces a new requirement for grid sysadmins.

OTOH, using the "grid location" means we'll be embedding that for the next decade.  A new system is the best chance to get out of old habits.

Thoughts?

docs/installation/htcondor-ce.md
brianhlin: How's it look now?

config/01-ce-auth-defaults.conf
```diff
@@ -59,9 +59,11 @@ GSI_DAEMON_KEY  = /etc/grid-security/hostkey.pem
 GSS_ASSIST_GRIDMAP_CACHE_EXPIRATION = 30*$(MINUTE)
 
 # SSL settings
-AUTH_SSL_SERVER_CAFILE=/etc/pki/tls/certs/ca-bundle.crt
-AUTH_SSL_SERVER_CERTFILE=/etc/grid-security/condor/hostcert.pem
-AUTH_SSL_SERVER_KEYFILE=/etc/grid-security/condor/hostkey.pem
+# TODO: These are the new defaults in 8.9 so these settings can be
+# removed when we no longer need to support HTCondor < 8.9
+AUTH_SSL_SERVER_CAFILE = /etc/pki/tls/certs/ca-bundle.crt
```
matyasselmeci: Try wrapping these in
```
if version <= 8.8
...
endif
```
(version comparison here is unusual because `version <= 8.8` is true for 8.8.X for any X)

brianhlin: I'm not sure if this buys us anything since we'll still want to remove this config when we stop supporting 8.9. An if block would make us sensitive to (unlikely) changes to the HTCondor default, too.

PR #256, opened 2019-10-07
Remove unnecessary documentation references to 'users.htcondor.org'

brianhlin:
Mapped HTCondor principles without an '@' suffix automatically have
'@users.htcondor.org' appended since it's the default
UID_DOMAIN. Exposing this to users has lead to confusion where they
replace 'htcondor.org' with their own domain, resulting in failure to
submit jobs.

### DIFF
--- a/config/condor_mapfile
+++ b/config/condor_mapfile
@@ -10,17 +10,17 @@
 # with '\/') with the distinguished name of the incoming user certificate and
 # the unix account under which the job should run, respectively:
 #
-# GSI "^<DISTINGUISHED NAME>$" <USERNAME>@users.htcondor.org
+# GSI "^<DISTINGUISHED NAME>$" <USERNAME>
 
 # VOMS attributes can also be used for mapping:
 #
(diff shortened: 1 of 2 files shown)


matyasselmeci:
_Is_ it the default UID_DOMAIN? I thought that was FULL_HOSTNAME. Or is the CE different?

brianhlin:
Yeah, it's the default for the CE: https://github.com/htcondor/htcondor-ce/blob/0bd02bf5516c5aa8e4eade8f06e0a7cda6c95704/config/01-ce-auth-defaults.conf#L23

The CE makes use of three UID domains that map to three effective authorization levels:

- `users.htcondor.org` for job submission
- `daemon.htcondor.org` for interaction between daemons and administration
- `unmapped.htcondor.org` for clients that fail to authenticate

matyasselmeci:
Does it have to look like an FQDN? Or can it be `@user`, `@daemon`, `@unmapped`?

brianhlin:
I'm not sure but I've always seen it as an FQDN and I think that's what users would expect since this shows up in accounting groups

matyasselmeci:
Well it's probably too late now but if it didn't look like an FQDN perhaps people wouldn't mistake it for an FQDN...

PR #270, opened 2019-10-23
Allow users to specify pilot job environments in a non-OSG specific way (SOFTWARE-3871)

brianhlin:
This PR adds two methods for users to specify pilot environment variables: setting `CONDORCE_PILOT_JOB_ENV` in the configuration or `set_default_pilot_job_env` in their router configuration. The former would be useful for setting static env vars or env vars based on other configuration values. The latter would be useful if they want to use ClassAd expressions or set vars based on route/job attributes. Both of these methods expect the "new environment" format as documented in the `condor_submit` man page. Env vars specified in this way will override env vars specified in the OSG env files or the original job's environment attribute.

This PR also adds a `DISABLE_PILOT_ADS` config knob that disables setting the `CondorCECollectorHost` env var, which can be used by pilots to advertise their startd's back to the site CE collector. The name is different from the `DISABLE_GLIDEIN_ADS` env var that users can set in `/etc/sysconfig/condor-ce` to be less GlideinWMS specific. The latter is also an undocumented feature so I think it's ok to have different names.

I've tested the changes by setting variables in `CONDORCE_PILOT_JOB_ENV`, `set_default_pilot_job_env`, adding `export DISABLE_GLIDEIN_ADS=True` in `/etc/sysconfig/condor-ce`, and setting `DISABLE_PILOT_ADS = True` in the HTCondor-CE config. So the CI should pass :crossed_fingers: 

After we get this merged, I'll cherry pick the changes and submit a PR for v3 as well. Also @matyasselmeci I'd like to see `osg-configure` updated to use `CONDORCE_PILOT_JOB_ENV` instead of the OSG env files so we could deprecate and retire the latter


brianhlin:
@matyasselmeci @djw8605 thoughts?

brianhlin:
TJ said the ClassAd expression looked fine to him

PR #278, opened 2019-11-08
Add traceback to failed_ce's for visibility

djw8605:


djw8605:
Inserting the full add into the `failed_ces` caused further failures.  In some cases, the ad was not JSON serializable, which caused it to fail when outputting the full object.

PR #297, opened 2019-12-19
Add SSL to the end of client auth methods (SOFTWARE-3939)

brianhlin:
In HTCondor 8.9, SSL is automatically added to the AuthMethodList on
the server side so we don't need an accompanying change on the server side.


brianhlin:
> The server side must currently be working because we're not touching the defaults there.

We do set `SEC_DEFAULT_AUTHENTICATION_METHODS` here: https://github.com/htcondor/htcondor-ce/blob/master/config/01-ce-auth-defaults.conf#L49. In my testing, SSL was added to the server-side auth method list no matter what and @zmiller verified this in the code. Perhaps we should explicitly add `SSL` to the default auth method list for clarity.

PR #298, opened 2019-12-26
Enable submission to HTCondor-CE via token exchange

bbockelm:
This PR enables the submission to a HTCondor-CE, using SciTokens, without a host certificate.

To do this, we first enable the TOKEN authentication method for the HTCondor-CE and automatically creates a pool password when the package is installed.

Then, this adds the `condor_ce_scitoken_exchange` script that allows someone with a remote identity at the collector to exchange a valid SciToken for a CE-specific TOKEN.

The resulting `TOKEN` is usable only at the specified CE for activities equivalent to those specified in the original SciToken.


CMakeLists.txt
```diff
@@ -152,20 +154,35 @@ set(HTCONDORCE_DIRS ${STATE_INSTALL_DIR}/run/condor-ce ${STATE_INSTALL_DIR}/log/
...
 
 foreach(sticky_dir IN ITEMS ${STICKY_HTCONDORCE_DIRS})
   install(CODE "execute_process(COMMAND chmod 1777 ${sticky_dir})")
 endforeach(sticky_dir)
 
+set(PRIVATE_HTCONDORCE_DIRS ${SYSCONF_INSTALL_DIR}/condor-ce/passwords.d
+  ${SYSCONF_INSTALL_DIR}/condor-ce/tokens.d)
+
+foreach(private_dir IN ITEMS ${PRIVATE_HTCONDORCE_DIRS})
+  install(CODE "execute_process(COMMAND chmod 0700 ${private_dir})")
```
matyasselmeci: Doesn't the `condor` user need to have access to passwords.d and/or tokens.d?

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
...
+    token_dir = htcondor.param.get('SEC_TOKEN_DIRECTORY')
+    if not token_dir:
+        if os.geteuid() == 0:
+            token_dir = htcondor.param['SEC_TOKEN_SYSTEM_DIRECTORY']
+        else:
+            token_dir = os.path.expanduser("~/.condor-ce/tokens.d")
+    try:
+        os.makedirs(token_dir)
+    except FileExistsError:
+            pass
```
matyasselmeci:
```suggestion
        pass
```

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
...
+    info = contents.split(".")
+    if len(info) != 3:
+        return False
+
+    payload = info[1]
+    while len(payload) % 4 != 0:
+        payload += "="
+
+    payload_contents = base64.b64decode(payload)
+    payload_obj = json.loads(payload_contents)
```
matyasselmeci:
```suggestion
    try:
        payload_obj = json.loads(payload_contents)
    except json.JSONDecodeError:
        return False
```

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
...
+def token_is_valid(contents):
+    info = contents.split(".")
+    if len(info) != 3:
+        return False
+
+    payload = info[1]
+    while len(payload) % 4 != 0:
+        payload += "="
+
+    payload_contents = base64.b64decode(payload)
```
matyasselmeci:
```suggestion
    try:
        payload_contents = base64.b64decode(payload)
    except binascii.Error:
        return False
```

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/python3
+
+import argparse
+import base64
+import json
```
matyasselmeci:
```suggestion
import binascii
import json
```

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
...
+    fetch_env.setdefault('_condor_SEC_TOKEN_DIRECTORY', token_dir)
+    args = ['condor_token_fetch', '-pool', pool, '-name', name,
+        "-lifetime", "3600", '-remote', '-authz', 'READ']
+    if debug:
+        args.append('-debug')
+    result = subprocess.run(args, stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE, env=fetch_env)
+    if result.returncode:
+        print("condor_token_fetch failed with error %d:" % result.returncode)
+        print(result.stderr.decode('ascii'), file=sys.stderr, end='')
```
matyasselmeci: Is the result of condor_token_fetch guaranteed to only contain ASCII characters?

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
...
+    exchange_env.setdefault('_condor_SEC_TOKEN_DIRECTORY', token_dir)
+    args = ['condor_scitoken_exchange', '-scitoken', scitoken_fname,
+        "-pool", pool, "-name", name]
+    if debug:
+        args.append("-debug")
+    result = subprocess.run(args, stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE, env=exchange_env)
+    if result.returncode:
+        print("condor_scitoken_exchange failed with error %d:" % result.returncode)
+        print(result.stderr.decode('ascii'), file=sys.stderr, end='')
```
matyasselmeci: Is the result of condor_scitoken_exchange guaranteed to only return ASCII characters?

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,132 @@
...
+            print("No pre-existing token found -- will acquire a new one from %s.\n" % pool, file=sys.stderr)
+        contents = acquire_ce_token(pool, args.name, token_dir, debug=args.debug)
+        with open(exchange_token_name, "w") as fp:
+            fp.write(contents.decode('ascii') + "\n")
+
+    result = exchange_scitoken(pool, args.name, args.scitoken, token_dir, debug=args.debug)
+    if result:
+        print(result.decode('ascii'), end="")
+
+    sys.exit(1)
```
matyasselmeci: Why 1?

rpm/htcondor-ce.spec
```diff
@@ -228,6 +228,11 @@ install -m 0755 -d -p $RPM_BUILD_ROOT/%{_sysconfdir}/condor-ce/bosco_override
 
 %post
 /bin/systemctl daemon-reload >/dev/null 2>&1 || :
+
+if [ ! -e /etc/condor-ce/passwords.d/POOL ]; then
+    %{_datadir}/condor-ce/condor_ce_create_password >/dev/null 2>&1 || ::
```
matyasselmeci:
```suggestion
    %{_datadir}/condor-ce/condor_ce_create_password >/dev/null 2>&1 || :
```

src/condor_ce_scitoken_exchange
bbockelm: Yup!  It's guaranteed to be base64-encoded.  I'm OK with this code throwing a nasty exception in the case of a hacked-up remote condor daemon.

src/condor_ce_scitoken_exchange
bbockelm: Yup!  Same as above, it's base64-encoded and would require a custom server compile to return something else (in which case, I'm OK with an inscrutable exception message).

CMakeLists.txt
bbockelm: No and yes.  `passwords.d` should be `root:root` and `0700` while `tokens.d` should be `condor:condor` and `0700`.  I _think_ that's what I have above.

src/condor_ce_scitoken_exchange
bbockelm: Sigh.  Leftover from an old refactoring.  I'll fix it.

matyasselmeci:
I think it's fine; @brianhlin , wanna take a look?

config/01-ce-auth-defaults.conf
```diff
@@ -68,6 +68,13 @@ AUTH_SSL_SERVER_KEYFILE = /etc/pki/tls/private/localhost.key
 # SciTokens settings
 SCITOKENS_SERVER_AUDIENCE=$(COLLECTOR_HOST)
 
+# Allow READ-level TOKEN authorizations to be generated by the collector.
```
brianhlin: - If this comment is true, we should move this configuration to https://github.com/htcondor/htcondor-ce/blob/master/config/04-ce-collector-auth.conf
- If it's needed for both the CE and the central collector, we should move it to https://github.com/htcondor/htcondor-ce/blob/master/config/01-common-auth-defaults.conf
- Otherwise if it's for a site CE's collector, this comment should be clarified

config/01-ce-auth-defaults.conf
```diff
@@ -68,6 +68,13 @@ AUTH_SSL_SERVER_KEYFILE = /etc/pki/tls/private/localhost.key
...
+SEC_ENABLE_IMPERSONATION_TOKENS=true
+SEC_IMPERSONATION_TOKEN_LIMITS=READ
+
+# Pool password directory for the CE.
+SEC_PASSWORD_DIRECTORY = /etc/condor-ce/passwords.d
```
brianhlin: If I understand it correctly, the central collectors and the CEs themselves need this dir. In that case, this should be moved to https://github.com/htcondor/htcondor-ce/blob/master/config/01-common-auth-defaults.conf

rpm/htcondor-ce.spec
```diff
@@ -290,6 +295,8 @@ install -m 0755 -d -p $RPM_BUILD_ROOT/%{_sysconfdir}/condor-ce/bosco_override
 %attr(-,condor,condor) %dir %{_localstatedir}/lib/condor-ce/execute
 %attr(-,condor,condor) %dir %{_localstatedir}/lock/condor-ce
 %attr(1777,condor,condor) %dir %{_localstatedir}/lock/condor-ce/user
+%attr(0700,root,root) %dir %{_sysconfdir}/condor-ce/passwords.d
+%attr(0700,condor,condor) %dir %{_sysconfdir}/condor-ce/tokens.d
```
brianhlin: Same as above: if both the central collectors and CEs need these dirs, this should go under `%files client`

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+
+import htcondor
+
+
+DEFAULT_COLLECTOR_PORT="9619"
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Exchange a SciToken for a HTCondor token at a remote CE")
+    parser.add_argument("-pool", help="Remote collector to query", metavar="coll",
+        default="collector.opensciencegrid.org")
```
brianhlin: I was thinking about this a bit more and this means that for this to work consistently, we need the collector1 <-> collector2 ad sharing feature implemented in HTCondor, then the HTCondor-CE central collector config.

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+    parser.add_argument("-name", help="CE name to exchange SciToken with", metavar="name",
+        required=True)
+    parser.add_argument("-scitoken", help="File containing SciToken to exchange", metavar="token_file",
+        required=True)
+    parser.add_argument("-debug", action="store_true", help="Enable debug mode")
```
brianhlin:
```suggestion
    parser.add_argument("-d", "--debug", action="store_true", help="Enable debug mode")
```
Let's not repeat the sins of our forefathers

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+
+os.environ.setdefault("CONDOR_CONFIG", "/etc/condor-ce/condor_config")
+
+import htcondor
+
+
+DEFAULT_COLLECTOR_PORT="9619"
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Exchange a SciToken for a HTCondor token at a remote CE")
```
brianhlin:
```suggestion
    parser = argparse.ArgumentParser(description="Create a token based off of a SciToken for authentication with a remote CE")
```
Exchange makes it sound like a trade

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+    parser.add_argument("-pool", help="Remote collector to query", metavar="coll",
```
brianhlin:
```suggestion
    parser.add_argument("-p", "--pool", help="Remote collector to query", metavar="coll",
```

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+        default="collector.opensciencegrid.org")
+    parser.add_argument("-name", help="CE name to exchange SciToken with", metavar="name",
```
brianhlin:
```suggestion
    parser.add_argument("-n", "--name", help="CE name to exchange SciToken with", metavar="name",
```

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+        required=True)
+    parser.add_argument("-scitoken", help="File containing SciToken to exchange", metavar="token_file",
```
brianhlin:
```suggestion
    parser.add_argument("-s", "--scitoken", help="File containing SciToken to exchange", metavar="token_file",
```

config/01-ce-auth-defaults.conf
bbockelm: The comment's right - maybe the English is just not clear.

That's a CE setting which allows the collector to generate TOKENs.  Will tweak.

src/condor_ce_scitoken_exchange
bbockelm: I'm not sure I follow here -- this should work with either collector just as well as it does today.

matyasselmeci: Why do we have the OSG collector as the default collector? I thought this PR was not supposed to be OSG-specific.

If you don't want to make everyone specify `-pool` then there should be a condor config knob for the default collector.

brianhlin: > I'm not sure I follow here -- this should work with either collector just as well as it does today.

If a CE has only successfully advertised to one of the collectors and the query gets DNS round-robined to the other collector, the command will fail. Admittedly, I think we've only seen this scenario come up once so I don't think it should hold up the PR but I thought it was worth highlighting.

> Why do we have the OSG collector as the default collector? I thought this PR was not supposed to be OSG-specific.
> 
> If you don't want to make everyone specify `-pool` then there should be a condor config knob for the default collector.

We also have this problem in [condor_ce_info_status](https://github.com/htcondor/htcondor-ce/blob/master/src/condor_ce_info_status#L3) and I've been trying to think of a good default. A config knob (`CONDORCE_CENTRAL_COLLECTOR`?) is a good idea but I'm not sure if we should leave it undefined and have the tools that depend on it fail with a message or if defaulting to the OSG collector would be sufficient. Regardless, I think we should address this in a subsequent PR.

djw8605: OSG's collectors do not run in DNS round-robin.

config/01-ce-auth-defaults.conf
bbockelm: Done.

rpm/htcondor-ce.spec
bbockelm: Done.

src/condor_ce_scitoken_exchange
bbockelm: Ok - if I'm reading this right, it's an item to note but not an action item for this PR.  Resolving.

bbockelm:
@brianhlin - I think I have all the comments resolved from the last few touch-up commits.  Please re-review.

PR #299, opened 2019-12-31
Implement CE registration app (SOFTWARE-4089)

bbockelm:
This implements a small CE webapp that allows sysadmins to "register" their CEs with the OSG (really, just approve token requests coming from their CE.

Here's a screenshot of the code page:
image

You don't see it, but that URL forced me to first login via CILogon.

The idea is that an admin would invoke `condor_ce_register` (this CLI is to be written...), get prompted to open https://registry.opensciencegrid.org/code in a browser window, and approve the request after logging in via CILogon.  Once approved, the CE will automatically fetch the token and install it locally.

Behind the scenes, CILogon would return the OSGID (e.g., OSG10001) as part of the login process and that ID would be associated already in topology with the CE hostname.  Provided that the requested CE identity matches the OSG topology registration, we know that this is the valid admin of the CE and we can safely approve the request.

The end-result is that the CE can now advertise with collector.opensciencegrid.org using only the approval from the authenticated, registered sysadmin.  No host certificate needed - this bootstraps the auth from scratch!

Why is this useful?  When combined with #298, the OSG collector can mediate the factory's exchange of a SciToken for a CE token, allowing the factory to authenticate with the CE _as the VO_ via the TOKEN method.  The TOKEN method, unlike SCITOKEN or GSI, does not require the CE to have a public DNS name, public IP address (CE only needs outgoing connectivity to the OSG collector), or host certificate.


bbockelm:
Note that testing this requires applying both https://github.com/htcondor/htcondor/pull/94 and https://github.com/htcondor/htcondor/pull/92 to the `condor_token_request_list` client.

The Flask webapp needs to be run behind Apache with `mod_auth_openidc`.  Here's my config:
```
OIDCProviderMetadataURL https://cilogon.org/.well-known/openid-configuration
OIDCClientID  cilogon:/client_id/ABCABC
OIDCClientSecret  XYZXYZ
OIDCCryptoPassphrase 123123

# OIDCRedirectURI is a vanity URL that must point to a path protected by this module but must NOT point to any content
OIDCRedirectURI https://collector.example.com/registry/callback

<Location /registry/>
   AuthType openid-connect
   Require valid-user
</Location>

WSGIDaemonProcess     CERegistry user=condorce_webapp group=condorce_webapp processes=2 threads=25 display-name=CERegistry
WSGIProcessGroup      CERegistry
WSGIScriptAlias       /registry /var/www/wsgi-scripts/condor-ce/condor-ce-registry.wsgi

# Recommendation from upstream mod_wsgi developers.
LogLevel info

# Required for wsgi-scripts directory to allow executing WSGI.
<Directory /var/www/wsgi-scripts>
  Require all granted
</Directory>
```

where the WSGI file is something like this:

```
from htcondorce.registry import create_app
application = create_app()
```

Locally I have modifications to set config variables and tinker with the `PYTHONPATH`; that shouldn't be needed in production.

We can probably try to incorporate these two files into the packaging once we get feedback from @djw8605.

src/condor_ce_scitoken_exchange
```diff
@@ -0,0 +1,142 @@
...
+        required=True)
+    parser.add_argument("-scitoken", help="File containing SciToken to exchange", metavar="token_file",
+        required=True)
+    parser.add_argument("-debug", action="store_true", help="Enable debug mode")
+    return parser.parse_args()
+
+
+def acquire_ce_token(pool, name, token_dir, debug=False):
+    fetch_env = dict(os.environ)
+    fetch_env.setdefault('_condor_SEC_TOKEN_DIRECTORY', token_dir)
```
matyasselmeci: You should handle `_CONDOR_SEC_TOKEN_DIRECTORY` as well.

brianhlin:
@jthiltges I know that you've got Apache set up on the central collectors to forward requests to the CE View and Let's Encrypt. Would this additional app (intended to be run on the same host) work with your current setup?

jthiltges:
At present, I've been using nginx as the frontend. The most straightforward is probably to switch to Apache + mod_auth_openidc, which shouldn't be too hard. I haven't yet tried nginx + lua-resty-openidc.

bbockelm:
@jthiltges - coming back to this one.  Can you perhaps switch to Apache in preparation of testing out this work?

bbockelm:
@brianhlin - don't forget we need to do this in a follow-up:

https://github.com/htcondor/htcondor-ce/pull/299/files#diff-37fcce4498caaf8a23278227aec914d9R21

That is, we need to use topology to manage a map from CILogon ID to authorized CEs.

brianhlin:
@matyasselmeci could you take a look over my commits to this PR? I'm mostly concerned about  https://github.com/htcondor/htcondor-ce/pull/299/commits/b2c324363065cc088e663c1273fbb8e5d8294d52

matyasselmeci:
> @matyasselmeci could you take a look over my commits to this PR? I'm mostly concerned about [b2c3243](https://github.com/htcondor/htcondor-ce/commit/b2c324363065cc088e663c1273fbb8e5d8294d52)

I haven't tested them, but they seem fine.

brianhlin:
There's a bunch of refactoring to be done here but since @JoshKarpel has done some major reorganization in https://github.com/HTPhenotyping/registration, we'll wait on the refactor until that's settled before merging in relevant improvements.

brianhlin:
CI passed, merging

PR #300, opened 2020-01-13
Add support for SSL auth for advertising to the central collector (SOFTWARE-3939, SOFTWARE-3940)

brianhlin:
I'm on the fence about which suffix we should map SSL clients of the central collector: `@unmapped.htcondor.org` or `@users.htcondor.org`. I settled on the latter since only allowing SSL for advertising. If we went with mapping to `@unmapped.htcondor.org` instead, we'd have to update the schedd `COLLECTOR_REQUIREMENTS` to accept that suffix and I don't think it really buys us anything.


djw8605:
GSI -> unmapped.htcondor.org
SSL -> users.htcondor.org

Should it be consistent?

brianhlin:
Sure. I did that and loosened the `COLLECTOR_REQUIREMENTS` to remove any UID namespace requirements. I could change it so that we accept `users.htcondor.org` OR `unmapped.htcondor.org` instead of accepting any UID namespace but I'm not sure that matters since we verify that the hostname matches between the `Name` and `AuthenticatedIdentity` attributes

brianhlin:
@djw8605 could I get a re-review?

PR #306, opened 2020-02-24
Add options for specifying job route order in new versions of HTCondor (stable)

brianhlin:
If this look ok, I'll make another PR for master


brianhlin:
ping @matyasselmeci

PR #311, opened 2020-03-13
Adding Benchmark parameters

sartiran:
Added the GLUE2Benchmark parameters for each GLUE2Resource. They are taken from the configuration parameters: 

```
HTCONDORCE_SpecValue_<spec_type> = <spec_value>
```


contrib/bdii/htcondor-ce-provider
```diff
@@ -221,11 +230,28 @@ def main():
...
             instances=instances,
         ))
+
+
+        # Get from the configuration types and values of the benchmarks
+        spec_types = htcondor.param.get('HTCONDORCE_SpecTypes').split(',') or []
+        for spec_type in spec_types:
+
+            spec_type = spec_type.strip()
+            spec_value = htcondor.param.get('HTCONDORCE_SpecValue_%s' % spec_type.replace('-', '_'))
```
brianhlin: Instead of separate config vars for the different spec types, maybe we should use a single variable whose value is a classad:

```
# tail -n 1 /etc/condor/config.d/99-local.conf
HTCONDORCE_SPEC = [ spec1 = "val1"; spec2 = "val2"; spec3 = "val3" ]
```

And parsing that in Python:

```
>>> import classad
>>> import htcondor
>>> specs = htcondor.param.get('HTCONDORCE_SPEC')
>>> ad = classad.ClassAd(specs)
>>> ad
[ spec1 = "val1"; spec2 = "val2"; spec3 = "val3" ]
>>> for spec, val in ad.items():
...     print spec, val
spec1 val1
spec2 val2
spec3 val3
```

I think this would be easier for an admin to sort out

brianhlin:
Hi @sartiran have you had the chance to take a look at my comments above?

sartiran:
Hi @brianhlin , not yet sorry, I'll have a look later today and try to fix things by tomorrow.

sartiran:
HI Brian, I've commited some changes that should answer your comments.

PR #314, opened 2020-03-19
Fix central collector SSL advertising (SOFTWARE-3939); remove config generator

brianhlin:
We enforce host identities via `COLLECTOR_REQUIREMENTS` so we don't need the whitelist generated by the config generator. We had planned to drop it in this ticket but forgot: https://opensciencegrid.atlassian.net/browse/SOFTWARE-1790.

I tested these changes with a test collector + CE and verified that the CE advertising worked with SSL.


brianhlin:
CI tests completed successfully but the Travis/GitHub connection appears to be having issues. Merging

PR #315, opened 2020-03-24
Add gsi@unmapped to UNMAPPED_USERS (SOTWARE-4039)

brianhlin:
When GSS_ASSIST_GRIDMAP fails, it results in the authenticated
identity of "gsi@unmapped". Additionally, the failed mappings do not
fall through to other GSI lines in the mapfile so the second GSI
mapfile lines were unnecessary here.

### DIFF
--- a/config/01-ce-auth-defaults.conf
+++ b/config/01-ce-auth-defaults.conf
@@ -23,9 +23,9 @@ FRIENDLY_DAEMONS = condor@daemon.htcondor.org/$(FULL_HOSTNAME) condor@child/$(FU
 UID_DOMAIN = users.htcondor.org
 USERS = *@users.htcondor.org
 
-# Unmapped users who authenticated with GSI; they are allowed to advertise
+# Unmapped users who authenticated with GSI or SSL; they are allowed to advertise
 # to the collector, but that is all.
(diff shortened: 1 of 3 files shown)


matyasselmeci:
Doesn't it fall through to anonymous@claimtobe if GSI and SSL fail?

brianhlin:
Nope, this is what happens:

1. Server and client negotiate an auth method
1. Server uses all the lines in the mapfile corresponding to the chosen auth method. If the chosen auth method doesn't have any lines in the mapfile, I believe there are default mapped IDs.
1. Staring with the top line, the server attempts to match the client's credentials to the second field of each line

So in our case, every single GSI-negotiated mapping will match to the first line in the mapfile since the regex is `(.*)`. If the GSI callout (i.e., `GSS_ASSIST_GRIDMAP`) fails, that results in the mapped identity of `gsi@unmapped` and the mapfile is not consulted again.

matyasselmeci:
OK, that makes sense. Yeah, putting that info in the commit message would be helpful -- condor auth is confusing :-)

I'll approve although Travis appears broken at the moment...

brianhlin:
Force push was just for amending the commit message. Merging

PR #318, opened 2020-04-16
Identify misconfigured routes and fix bad ClassAd detection (SOFTWARE-3896, SOFTWARE-4052)

brianhlin:
Check out c39c30d45b90fed8878ddff3235a9db3bfc1def2 for the bulk of the changes since the commit adding unit tests makes a big bulk indentation change.

This change is for `stable` and I'll cherry-pick the changes to `master` once they're approved. Closes #233

### DIFF
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -62,7 +62,12 @@ def main():
             error("Missing required %s configuration value" % attr)
 
         # store the ads (iterating through ClassAdStringIterator consumes them)
-        parsed_jr_ads[attr] = list(classad.parseAds(config_val))
+        try:
+            parsed_jr_ads[attr] = list(classad.parseAds(config_val))
+        except ValueError:
+            # We shouldn't ever get here since classad.parseAds() only raises ValueError when it's given
+            # non-string/non-file output and htcondor.param shouldn't contain such values
+            error("Failed to parse %s configuration value" % attr)
 
         # If JRD or JRE can't be parsed, the job router can't function
         if not parsed_jr_ads[attr]:
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -31,7 +31,7 @@ def malformed_entries(entries_config):
     """Find all unparseable router entries based on the raw JOB_ROUTER_ENTRIES configuration
     """
     unparsed_names = [x.replace('"', '')
-                      for x in re.findall(r'name\s*=\s*["\'](\w+)["\']',
+                      for x in re.findall(r'''name\s*=\s*["'](\w+)["']''',
                                           entries_config,
                                           re.IGNORECASE)]
     parsed_names = parsed_route_names(entries_config)
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -17,26 +17,22 @@ def warn(msg):
     print "WARNING: " + msg
 
 
-def condorpp(param):
-    syntax = [';', '[', ']']
-    out = "".join(param.split())  # strip all whitespace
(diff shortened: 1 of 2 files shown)


brianhlin:
Oops, force push was because I left a WIP commit in there that's no longer relevant

src/verify_ce_config.py
```diff
@@ -10,86 +10,135 @@
...
+    """Return names of job routes that can be parsed as proper ClassAds
+    """
+    return [x['Name'] for x in classad.parseAds(entries_config)]
+
+
+def malformed_entries(entries_config):
+    """Find all unparseable router entries based on the raw JOB_ROUTER_ENTRIES configuration
+    """
+    unparsed_names = [x.replace('"', '')
+                      for x in re.findall(r'name\s*=\s*["\'](\w+)["\']',
```
matyasselmeci, line 34:
```suggestion
                      for x in re.findall(r'''name\s*=\s*["'](\w+)["']''',
```
The `r` causes the backslash to get in there literally, even when you're trying to quote the quote character.

src/verify_ce_config.py
```diff
@@ -10,86 +10,135 @@
...
 def error(msg):
+    """Exit with error 'msg'
+    """
     sys.exit("ERROR: " + msg)
 
 
 def warn(msg):
+    """Print warning 'msg'
+    """
     print "WARNING: " + msg
```
matyasselmeci, line 21:
```suggestion
    print("WARNING: " + msg)
```

src/verify_ce_config.py
```diff
@@ -10,86 +10,135 @@
...
 
 
+def parsed_route_names(entries_config):
+    """Return names of job routes that can be parsed as proper ClassAds
+    """
+    return [x['Name'] for x in classad.parseAds(entries_config)]
```
matyasselmeci, line 27: You might want to do something with whatever exceptions classad.parseAds() might throw on unparseable ads.

src/verify_ce_config.py
```diff
@@ -10,86 +10,135 @@
...
+                         % ', '.join(missing_route_def))
+
+    # Find all eval_set_ attributes in the JOB_ROUTER_DEFAULTS
+    eval_set_defaults = set([x.lstrip('eval_') for x in parsed_jr_ads['JOB_ROUTER_DEFAULTS'][0].keys()
+                             if x.startswith('eval_set_')])
+
+    # Find all default_ attributes used in expressions in the JOB_ROUTER_DEFAULTS
+    default_attr = set([re.sub(r'.*(default_\w*).*', 'eval_set_\\1', str(x))
+                        for x in parsed_jr_ads['JOB_ROUTER_DEFAULTS'][0].values()
+                        if isinstance(x, classad.ExprTree) and str(x).find('default_') != -1])
```
matyasselmeci, line 96:
```suggestion
                        if isinstance(x, classad.ExprTree) and "default_" in str(x)])
```

brianhlin: Much better, thanks!

src/verify_ce_config.py
brianhlin: Unfortunately, strings that have unparseable ads in them don't raise any sort of exception. The unparseable ads are just ignored.

However, `ValueError` is raised if you pass `classad.parseAds` a non-string value so we could catch that but I'm not sure what the proper behavior would be other than re-raising the exception.

matyasselmeci: It should be OK to catch it elsewhere in the program -- I'd just like to make sure there's a nice error message instead of a stack trace.

PR #329, opened 2020-06-02
index.html template fixes

matyasselmeci:
Mostly whitespace but I noticed a few HTML5 bugs.


djw8605:
Looks good.  Looks like mostly whitespace changes.

Issue #323, opened 2020-04-29
Inconsistent APEL permissions/user


maxfischer2781:
It seems the ``htcondor-ce-apel`` is written with a regular cronjob (as ``root``) in mind, not a Schedd_Cron job (as ``condor``). The standard permissions of ``/var/lib/condor-ce/apel/`` and ``/var/log/apel/`` prevent the ``condor`` user from writing to them, causing the Schedd_Cron job to fail.

I'm not sure whether permissions of the folders should be changed, or the job should be run as root. Only the ``/var/lib/condor-ce/apel/`` folder is created by ``htcondor-ce-apel``.

brianhlin:
Hi @maxfischer2781, thanks for the report! I'm curious, if you change the `logfile` in `/etc/apel/logging.cfg` and `/etc/apel/client.cfg` to a path that the `condor` user can write to, does that solve the issue with `/var/log/apel/`?

maxfischer2781:
Yes, changing the logfile to a writeable location fixes the problem. The paths in ``/etc/apel/client.cfg`` (default ``logfile = /var/log/apel/client.log``), ``/etc/apel/parser.cfg``(default ``logfile = /var/log/apelparser.log``), and ``/etc/apel/sender.cfg`` (default ``logfile: /var/log/apel/ssmsend.log``) need changing.

maxfischer2781:
Just a heads-up that we've missed some other paths (due to tinkering with permissions) that APEL will attempts to write to by default as root. These are all below ``/var/spool/apel/``, ``/var/log/apel/``, plus ``/var/log/apelparser.log``. Subdirectories are created for the current user, e.g. ``/var/spool/apel/outgoing/`` as ``drwxr-xr-x``.

brianhlin:
Hi @maxfischer2781, I think we're going to have to abandon the idea of running the APEL tools as a non-root user via `SCHEDD_CRON`. There are just too many paths that we're having users change in their configurations -- the next version(s) of HTCondor-CE will have systemd units and timers and cron jobs for the APEL integration

maxfischer2781:
Thanks a lot, that seems viable. I'll work around the issue manually for now.

brianhlin:
Resolved by #330 and #334

PR #330, opened 2020-06-09
Add APEL systemd service and timer

brianhlin:
From #323, the APEL SchedD cron has not worked out due to various APEL directory permission issues. Replace it with a systemd service and timer.


matyasselmeci:
Would a cron job be "better" since it would work in a container?

brianhlin:
I was going to add a cron job for the `stable` branch for EL6 support but suppose I could add that here, too.

brianhlin:
Though it doesn't actually make sense to ship both for EL7. I think I'd rather have the cron for EL6 installs and container images can add their own cron.

PR #337, opened 2020-06-15
Separate out OSG-specific parts of CEView

matyasselmeci:
For SOFTWARE-3963. Two parts to this PR (I could make them separate PRs if you want):

- Move the "/json/agis-compat" endpoint to a 'plugin' that gets loaded at runtime ('k fine it's Python, everything gets loaded at runtime). This plugin only gets packaged in OSG builds.

- Allow configuring which attributes show up in the "Info" table on the main page. The UW build ships with a config that just shows CondorVersion and HTCondorCEVersion. The OSG build also shows the OSG_Resource, OSG_ResourceGroup, etc.

Known issue: the attributes in the HTCONDORCE_VIEW_INFO_TABLE_ATTRIB_\* are case-sensitive.


src/htcondorce/plugins/agis_json.py
```diff
@@ -0,0 +1,89 @@
...
+    results = agis_data(environ)
+
+    start_response(OK_STATUS, _headers('application/json'))
+
+    return [ json.dumps(results) ]
+
+
+urls = [
+    (re.compile(r'^json/+agis-compat$'), agis_json),
+]
```
brianhlin, line 89:
```suggestion
]

```

With this, can't we also drop `agis_data` from `src/htcondorce/web_utils.py`?

config/05-ce-view-table.osg.conf
```diff
@@ -0,0 +1,27 @@
...
+HTCONDORCE_VIEW_INFO_TABLE_LABEL_1  = OSG Resource
+HTCONDORCE_VIEW_INFO_TABLE_ATTRIB_1 = OSG_Resource
+HTCONDORCE_VIEW_INFO_TABLE_LABEL_2  = OSG Resource Group
+HTCONDORCE_VIEW_INFO_TABLE_ATTRIB_2 = OSG_ResourceGroup
+HTCONDORCE_VIEW_INFO_TABLE_LABEL_3  = Batch System
+HTCONDORCE_VIEW_INFO_TABLE_ATTRIB_3 = OSG_BatchSystems
+HTCONDORCE_VIEW_INFO_TABLE_LABEL_4  = HTCondor-CE version
+HTCONDORCE_VIEW_INFO_TABLE_ATTRIB_4 = HTCondorCEVersion
+HTCONDORCE_VIEW_INFO_TABLE_LABEL_5  = HTCondor version
+HTCONDORCE_VIEW_INFO_TABLE_ATTRIB_5 = CondorVersion
```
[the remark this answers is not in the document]
matyasselmeci: Agreed; I didn't have the time to do it now, but we can do that later.

rpm/htcondor-ce.spec
```diff
@@ -214,12 +219,19 @@ rm -f $RPM_BUILD_ROOT%{_datadir}/condor-ce/condor_batch.sh
 rm -f $RPM_BUILD_ROOT%{_datadir}/condor-ce/condor_ce_apel.sh
 rm -f $RPM_BUILD_ROOT%{_unitdir}/condor-ce-apel.service
 rm -f $RPM_BUILD_ROOT%{_unitdir}/condor-ce-apel.timer
+mv -f $RPM_BUILD_ROOT%{_sysconfdir}/condor-ce/config.d/05-ce-view-table.osg.conf \
+      $RPM_BUILD_ROOT%{_sysconfdir}/condor-ce/config.d/05-ce-view-table.conf
```
brianhlin, line 223: Since your last commit adds defaults to `/usr/share/`, let's move the OSG config there instead of `/etc/`

src/htcondorce/web.py
```diff
@@ -19,21 +21,56 @@
...
         ce_config = environ.get('htcondorce.config', '/etc/condor-ce/condor_config')
         htcondor = htcondorce.web_utils.check_htcondor()
+
+        plugins_dir = htcondor.param.get("HTCONDORCE_VIEW_PLUGINS_DIR", "/usr/share/condor-ce/ceview-plugins")
+        if os.path.isdir(plugins_dir):
+            for filename in sorted(os.listdir(plugins_dir)):
+                if not filename.endswith(".py"):
+                    continue
+                name = filename[:-3]
+                plugin = imp.load_source(name, os.path.join(plugins_dir, filename))
```
djw8605, line 69: The imp interface has been deprecated in favor of the importlib module. https://docs.python.org/3/library/importlib.html#module-importlib

In fact, I can't find the load_source function in the python 3's documentation of the old imp module.

matyasselmeci: Yeah, we'll have to fix that up when we port this to Python 3.

Issue #346, opened 2020-06-17
Various minor CEView UI issues


matyasselmeci:
- [ ] Main page: the "Pilots" and "Info" tables take a couple of seconds to load and look blank until then. Maybe add some "Loading..." text so the user doesn't think something's broken.
- [ ] [VOs page](https://collector1.opensciencegrid.org/vos) could also use the "Loading..." text, and is also blank if you don't have any VOs; maybe some text saying "you don't have any VOs" so the user doesn't think something's broken.

ColeBollig:
Change close type

PR #355, opened 2020-07-01
Various CEView polishing: Info table, RRD generation, ClassAd to JSON

jthiltges:
- Request both schedds and tableattribs at the same time, waiting until both requests are complete
- Fix the HTML tag nesting for the table

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -18,7 +18,10 @@
             $.get('/json/totals', got_totals);
             $.get('/json/pilots', got_pilots);
             if("${multice}" != "False") {
-                $.get('/json/schedds', got_schedds);
+                $.when(
+                    $.get('/json/schedds'),
+                    $.get('/json/tableattribs'),
+                ).then(function(a1, a2) { got_schedds(a1[0], a2[0]) });
             } else {
                 $.get('/json/schedd', got_info);
             }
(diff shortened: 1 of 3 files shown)


brianhlin:
Hey @jthiltges, are these just fixes that were needed for `collector-itb` or are they related to the issues that we're having with the production Central Collector?

jthiltges:
Hi Brian. These are related to collector-itb only. I'm still working on the odd response time issues. Would you prefer separate PRs for various proposed changes, or just commits on a single PR (with rebasing as needed)?

brianhlin:
A single rebased PR works for me!

jthiltges:
The JSON conversion changes should help with the slow loading mentioned in #346. From quick testing, it dropped the response time for /json/schedds from 6s to 0.4s.

For the RRD-to-PNG graph generation, we've got 13k tempfiles on the production collector at the moment that didn't get cleaned up. Moving it into memory should avoid this issue.

brianhlin:
@jthiltges do you have these changes deployed on the ITB collector already?

jthiltges:
Changes are running on ITB, yes. I also did a quick test on the ITB CEView, temporarily pointing it at the production collector, and the page also looked OK.

Places in this PR where there are most likely to be mistakes:
- The JQuery changes
- The ClassAd serialization

templates/index.html
```diff
@@ -51,32 +54,34 @@
...
-                info_table.append("<tr>");
-                info_table.append("<td>" + name + "</td>");
-                attribs.forEach(function(it) {
-                    info_table.append("<td>" + key[it.attrib] + "</td>");
-                });
-                info_table.append("</tr>");
-            }
-            info_table.append("</tbody>");
+          thead = $('<thead>');
+          row = $('<tr/>');
```
matyasselmeci, line 65: Uh does this work? `<tr>` is certainly not an empty tag.

jthiltges: My understanding is that append() puts data inside an element, rather than after.

The intent was to create a row element, then use append() to fill it with data cells.

matyasselmeci:
The ClassAd serialization doesn't work if you have an UNDEFINED in a nested classad or array. The preprocessing doesn't catch it, and json.dumps() thinks it's an int and renders it as `2`.

Try something like (untested):
```
def schedds(environ, start_response):
    ads = htcondorce.web_utils.get_schedd_ads(environ)
    results_str = "{"
    for ad in ads:
        if 'Name' not in ad:
            continue
        results_str += '"%s": %s,' % (ad['Name'], ad.printJson())
    results_str = results_str[:-1] + "}"  # replace trailing comma

    start_response(OK_STATUS, _headers('application/json'))

    return [ results_str ]
```

brianhlin:
Let's have a separate PR. @matyasselmeci could you create it and work with @jthiltges to get it tested on the collector-itb?

PR #357, opened 2020-07-01
Do not put removed jobs on hold

brianhlin:
Without excluding removed jobs, this hold expression can result in a
race condition where jobs are removed and immediately put back on hold
so that they remain in the queue and oscillate between these two states


matyasselmeci:
Honestly I think we shouldn't need this -- HTCondor shouldn't allow moving a job out of the removed state once it's in there...

brianhlin:
I think @tannenba asserted something similar but I'm not sure what other use cases there may be to allow for putting removed jobs back on hold.

matyasselmeci:
Nobody should have any. Removed state is temporary until the job is actually deleted from the queue, with no guarantee that anything would have time to run before that happens.

JaimeFrey:
We are considering changing HTCondor to not apply the hold policy expressions to removed jobs (#7723). I can foresee us also disallowing condor_hold to affect removed jobs. Putting removed jobs into held status was added for the grid universe for cases where we couldn't remove the downstream job and related resources. Back in the day, Miron was big on not leaving a mess on the machines we submit jobs to.

PR #359, opened 2020-07-06
CEView: Handle classad.Value.Undefined in ClassAd to JSON

jthiltges:
Address the issues from PR #355 with serializing HTCondor `Undefined`.

Unfortunately, `classad.Value.Undefined` is of type `int`, and `json.dumps()` converts it to "2". (The custom `JSONEncoder` class is only called for unknown types.) Instead use HTCondor JSON conversion, then back to a Python object.

The round-trip through JSON isn't great, but we want to concatenate multiple ClassAd objects into a single JSON object. This should avoid unexpected surprises that may arise from treating JSON as raw strings.

### DIFF
--- a/src/htcondorce/web.py
+++ b/src/htcondorce/web.py
@@ -79,23 +79,17 @@ def _headers(content_type):
             ('Cache-Control', 'max-age=60, public')]
 
 
-class ClassadEncoder(json.JSONEncoder):
-    def default(self, obj):
-        if isinstance(obj, classad.ExprTree):
-            return str(obj)
-        else:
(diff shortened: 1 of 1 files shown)


brianhlin:
@matyasselmeci does this fully address your concerns from #355?

matyasselmeci:
It's not perfect -- I think expressions inside nested ads and lists don't get evaluated -- but fully handling that is a very complicated problem that IMO we should punt on solving until we need it. I have some experimental code from Todd Miller but it only works with 8.9.

PR #371, opened 2020-07-27
Adjust APEL defaults of APEL_BATCH_HOST to lrms

maxfischer2781:
Fixes the default APEL setting for the BATCH_HOST. This previously pointed to the CE host instead of the LRMS host.

This change relies on ``JOB_ROUTER_SCHEDD2_POOL`` pointing at the LRMS collector, which seems to be always valid when the LRMS is HTCondor.

### DIFF
--- a/contrib/apelscripts/50-ce-apel-defaults.conf
+++ b/contrib/apelscripts/50-ce-apel-defaults.conf
@@ -8,7 +8,7 @@
 ###############################################################################
 
 APEL_CE_HOST = $(CONDOR_HOST)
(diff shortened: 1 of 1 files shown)


brianhlin:
@maxfischer2781 I think I was premature in accepting this PR. `APEL_BATCH_HOST` is used when constructing the APEL batch record, which previously combined the job ID with the CE host. Replacing the CE host with the central manager doesn't actually make sense here since jobs are not tied to the central manager but rather job IDs are unique for a given submit host (e.g. a CE host in front of an HTCondor batch system).

PR #372, opened 2020-07-27
Spec fixes for /etc/httpd/conf.d/htcondorce_registry.conf

jthiltges:
- Config file comments note `This file will NOT be overwritten on upgrade.`
- Config file shouldn't need to be executable

PR opened as draft as I haven't tested/rebuilt the package with the spec changes.


jthiltges:
Nothing else to note at the moment. Thank you for the review.

jthiltges:
Nothing else to note at the moment. Thank you for looking at it.

PR #373, opened 2020-07-27
Revert "Adjust APEL defaults of APEL_BATCH_HOST to lrms"

brianhlin:
This reverts commit 465221a84843c5d8ec4957c1f2dcee251e7cc643 from #371 

APEL_BATCH_HOST is used to construct the unique APEL batch record,
combining the local condor batch ID and the relevant host. Since job
IDs are unique to their submit hosts, this should be tied to the CE
host instead of the local central manager.

### DIFF
--- a/contrib/apelscripts/50-ce-apel-defaults.conf
+++ b/contrib/apelscripts/50-ce-apel-defaults.conf
@@ -8,7 +8,7 @@
 ###############################################################################
 
 APEL_CE_HOST = $(CONDOR_HOST)
-APEL_BATCH_HOST = $(JOB_ROUTER_SCHEDD2_POOL)
+APEL_BATCH_HOST = $(CONDOR_HOST)
 
 # The CE_ID is the CE, the port, and the batch system,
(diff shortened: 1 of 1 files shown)


matyasselmeci:
You should add a comment to explain why it's this way.

maxfischer2781:
I might be misunderstanding something, but I don't think the ID should be fixed this way. APEL needs to know the LRMS that ran the jobs. Using the Global Job Id from the local condor (i.e. LRMS Schedd) should work, doesn't it?

brianhlin:
Hi @maxfischer2781 we're not using the Global Job Id but rather the local job ID, so combining it with the LRMS submit host (i.e. the CE host) should be sufficient to identify the unique job.

When you say that APEL needs to know the LRMS that ran the jobs, do you mean it needs to know the name of the LRMS (e.g. condor, slurm, etc)?

maxfischer2781:
Sorry, I may have been jumping to conclusions here. Need to dig through APEL some more to understand what exactly it is doing with the information.

PR #382, opened 2020-08-14
Tenderize process_one_schedd (SOFTWARE-4183)

edquist:
Refactor this function a bit in prep for other SOFTWARE-4183 work.

The most scary looking part was the using of `.values()` for the job_key, which i don't think is actually guaranteed to be sorted.  The rest is mainly intended to make this function easier to work with.


brianhlin:
@edquist could you rebase this on master to make GitHub happy?

PR #381, opened 2020-08-13
collect stats for gpu jobs (SOFTWARE-4183)

edquist:
The first block of commits are cleanup and could probably stand alone.

There is still plumbing work to be done for the collected gpu info.

PR #383, opened 2020-08-20
add missing print >> redirect

edquist:
discovered randomly while debugging #381

### DIFF
--- a/src/condor_ce_host_network_check
+++ b/src/condor_ce_host_network_check
@@ -411,7 +411,7 @@ def main():
                 print >> sys.stderr, "Host network configuration not expected to work with HTCondor-CE."
                 sys.exit(1)
     else:
-        print sys.stderr, "WARNING: Unable to access certificate file at %s; skipping host certificate name check." % hostcert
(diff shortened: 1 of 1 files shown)


matyasselmeci:
It should be a separate PR, but we should be using Python 3-compatible syntax in our code, i.e. `from __future__ import print_function` at the top and `print("blah blah blahp", file=sys.stderr)`...

brianhlin:
@edquist could you cherry-pick this to the `stable` branch?

edquist:
> It should be a separate PR, but we should be using Python 3-compatible syntax in our code, i.e. `from __future__ import print_function` at the top and `print("blah blah blahp", file=sys.stderr)`...

Python3 support is going to be a large overhaul, i think, for htcondor-ce.  It contains a large amount of python2-specific code all over the place - so i wasn't going to pretend to make a difference in this fix  ;)

matyasselmeci:
Fair :)
It's pretty high on the prio list for RHEL 8 though, so I suspect we'll be getting to Python 3 fairly soon anyway.

Issue #376, opened 2020-07-31
Using real slots provided instead of detected (HT) cores for BDII


kreczko:
Currently, the htcondor-ce-provider uses the query
```python
coll.query(htcondor.AdTypes.Startd, 'State=!="Owner"',
                              ['Arch', 'OpSys', 'OpSysMajorVer', 'OpSysName',
                               'DetectedCpus', 'DetectedMemory', 'Machine'])
```
(see https://github.com/htcondor/htcondor-ce/blob/master/contrib/bdii/htcondor-ce-provider#L190)
to collect information about the cluster - which is based on detected (HT) cores.

The [GridPP HTCondorCE docs](https://www.gridpp.ac.uk/wiki/Example_Build_of_an_HTCondor-CE_Cluster#Appendix_1_-_Fix_the_BDII_provider_to_use_real_slots_provided.2C_not_just_detected_CPUS) recommend a hack to replace `'State=!="Owner"'` with `'SlotTypeID == 1'` in order to use partitionable slots instead.

Would it be against the current concept to make that change in the main distribution? If not, I would be happy to provide a PR for this.

brianhlin:
Hi @kreczko, I'm not sure that constraint would necessarily get you want you need since `SlotTypeID` is based on the slot types in the startd config and you could have more than one slot type. We could, however, update the query so that it only includes partitionable and static slots.

Are these core numbers supposed to indicate what is currently available in the cluster? If so, we can also make the `DetectedCpus` -> `TotalSlotCpus` change and make sure to ignore claimed slots.

maxfischer2781:
We're indeed running multiple partitionable slots of different type on some worker nodes. So actually checking for all proper slots instead of just the first type is required in our case.

Other than that, it seem reasonable to test for the available slots, not the physical cores.

kreczko:
> Are these core numbers supposed to indicate what is currently available in the cluster? If so, we can also make the `DetectedCpus` -> `TotalSlotCpus` change and make sure to ignore claimed slots.

Yes, `DETECTED_CPUS` might actually be better. We are trying to convey here how many cores are available for computing.
On most machines that's "cores excl. HT &leq; DETECTED_CPUS < cores including HT"

brianhlin:
Hi @kreczko I finally got the chance to take a look at this and I'm a little confused by your last comment. I was suggesting that we we only gather information for the unused capacity in a pool, not total capacity by counting all the unclaimed slots. This would look something like this:

```
[...]

PR #386, opened 2020-08-26
Avoid double-counting available resources

brianhlin:
1. Detected* attributes contain the total available resources on a
given machine
2. Dynamic slots are carved off of partitionable slots based on job
resource requests
2. Machines may have multiple slots

So to get the total number of resources available, we can just look at
the partitionable and static slots and sum the 'TotalSlot*' attributes
to get total pool CPUs and memory

TJ says this is only theoretically correct because of backfill behavior [1] but dealing with that will make the queries really heinous. I think this should be close enough.

[1]

> In theory, but since we lie about total cpus in order to do backfill on some machines,  what you actually want to do is grab TotalSlotCpus from p-slots, and ignore static slots unless that machine has no p-slots
> similarly, sum cpus of p-slots and igore static slots unless the machine has no p-slots.

### DIFF
--- a/contrib/bdii/htcondor-ce-provider
+++ b/contrib/bdii/htcondor-ce-provider
@@ -186,25 +186,25 @@ def main():
 
     # Query collector for the number of CPUs and batch system Collector ad
     coll = htcondor.Collector()
-    total_cores = {}
+    total_cores = 0
     topologies = {}
-    for classad in coll.query(htcondor.AdTypes.Startd, 'State=!="Owner"',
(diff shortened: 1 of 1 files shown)


brianhlin:
This PR will fix #376 as outlined but I'm a little unclear as to whether we want to count total resources or unused resources. If I understand it correctly, we should be counting total CPU/Mem for a pool and each slot should be considered a "resource" whereas previously each machine was considered a "resource".

@lfield what exactly does a resource LDIF represent?

brianhlin:
After some additional discussion in #376, we decided that this isn't necessary

PR #388, opened 2020-09-17
rename compute element -> entrypoint (stable) (SOFTWARE-4265)

edquist:


brianhlin (changes requested):
Same request here for reverting changes to links

brianhlin (approved):

PR #389, opened 2020-09-17
rename compute element -> entrypoint (master) (SOFTWARE-4265)

edquist:


brianhlin (changes requested):
Same thing here, let's avoid the link changes until we can get redirects from MkDocs

brianhlin (approved):
LGTM

PR #398, opened 2020-11-20
Fix CHTC repo non-HTCondor package installs (HTCONDOR-84)

brianhlin:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-84

Bonus removal of unnecessary `grid-certificates` OSG package requirement

### DIFF
--- a/rpm/htcondor-ce.spec
+++ b/rpm/htcondor-ce.spec
@@ -3,7 +3,7 @@
 
 Name: htcondor-ce
 Version: 4.4.1
-Release: 2%{?gitrev:.%{gitrev}git}%{?dist}
+Release: 3%{?gitrev:.%{gitrev}git}%{?dist}
 Summary: A framework to run HTCondor as a CE
 BuildArch: noarch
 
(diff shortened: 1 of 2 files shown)


matyasselmeci:
We still have GSI as one of the default auth methods, right? Is auth going to work without certificates, or will auth bail out due to GSI being unhappy? Non-OSG people aren't going to install osg-ce...

brianhlin:
Correct, we tell non-OSG users to install certificates separately since there isn't an equivalent `osg-certificates` package elsewhere since that's just not how they manage site CAs outside of the OSG: https://htcondor-ce.readthedocs.io/en/latest/installation/htcondor-ce/#before-starting

Plus, the requirement was only for OSG packaging so this has no effect for the non-OSG case.

matyasselmeci:
Looks like you'll need to install "python3-rpm" to get the osg tests to work. Dunno about the uw tests missing cagen.

rpm/htcondor-ce.spec
```diff
@@ -141,12 +126,8 @@ Requires: blahp
 Group: Applications/System
 Summary: Default routes for submission to Slurm
 Requires: %{name} = %{version}-%{release}
-
-%if 0%{?uw_build}
 Requires: /usr/libexec/condor/glite/bin/batch_gahp
```
timtheisen, line 129: I think you missed this one.

brianhlin: Good catch!

brianhlin:
@matyasselmeci since the CI is hung up for whatever reason and this is a straightforward, urgent fix, can you re-review? I'd rather not solve all the CI issues in this PR.

matyasselmeci (approved):

PR #405, opened 2021-01-26
Fixup collector stacktrace; add a UW build check for the CE collector

brianhlin:
Running OSG tests against 8.9 fixes the collector stacktrace since 8.9 provides better Python 3 support for the collector plugins.

Note that OSG CE View tests are still failing but I'll tackle that in a separate PR


timtheisen (approved):
LGTM

PR #406, opened 2021-01-26
Various CE View fixes (HTCONDOR-13)

brianhlin:
- The usual byte-string stuff
- Who knew that `Dict().keys()` now returns a `dict_keys` "view" object?

### DIFF
--- a/src/condor_ce_metric
+++ b/src/condor_ce_metric
@@ -1,7 +1,6 @@
 #!/usr/bin/python3
 
 import optparse
-import rrdtool
 from htcondorce import rrd, web_utils
 
 def parse_opts():
@@ -44,10 +43,10 @@ def main():
     environ = {'htcondorce.spool': spooldir}
 
     fname = rrd.check_rrd(environ, None, "metrics", opts.group, opts.name)
-    rrdtool.update(fname, "N:%s" % str(value))
+    rrd.update(fname, "N:%s" % str(value))
 
     fname = rrd.check_rrd(environ, opts.spoof.split(":")[-1], "metrics", opts.group, opts.name)
-    rrdtool.update(fname, "N:%s" % str(value))
+    rrd.update(fname, "N:%s" % str(value))
 
 
 if __name__ == '__main__':
--- a/src/condor_ce_metric
+++ b/src/condor_ce_metric
@@ -43,12 +43,11 @@ def main():
     environ = {'htcondorce.spool': spooldir}
 
     fname = rrd.check_rrd(environ, None, "metrics", opts.group, opts.name)
-    rrd.update(fname, "N:%s" % str(value))
+    rrd.update(fname, "N:%s" % value)
 
     fname = rrd.check_rrd(environ, opts.spoof.split(":")[-1], "metrics", opts.group, opts.name)
-    rrd.update(fname, "N:%s" % str(value))
+    rrd.update(fname, "N:%s" % value)
 
 
 if __name__ == '__main__':
     main()
-
--- a/src/htcondorce/rrd.py
+++ b/src/htcondorce/rrd.py
@@ -213,3 +213,11 @@ def graph(environ, host, plot, interval):
 
     return graph
 
+
+def update(fname, value):
+    try:
+        rrdtool.update(fname, value)
+    except NameError:
+        cmd = ['/usr/bin/rrdtool', 'update', fname, value]
+        proc = subprocess.Popen(cmd)
+        proc.communicate()  # wait until 'rrdtool update' completes
--- a/src/htcondorce/rrd.py
+++ b/src/htcondorce/rrd.py
@@ -6,7 +6,7 @@ import errno
 # Try using rrdtool Python bindings
 try:
     import rrdtool
-except ImportError:
+except ModuleNotFoundError:
     # If not available, fallback to shelling out to the CLI
     import subprocess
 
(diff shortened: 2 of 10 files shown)


edquist:
> Who knew that Dict().keys() now returns a dict_keys "view" object?

Welcome to python3  :)

src/htcondorce/rrd.py
```diff
@@ -6,7 +6,7 @@
 # Try using rrdtool Python bindings
 try:
     import rrdtool
-except ImportError:
+except ModuleNotFoundError:
```
edquist, line 9: Fine, but i think `ImportError` still works, since `ModuleNotFoundError` inherits from it

src/htcondorce/rrd.py
```diff
@@ -213,3 +213,11 @@ def graph(environ, host, plot, interval):
...
     return graph
 
+
+def update(fname, value):
+    try:
+        rrdtool.update(fname, value)
+    except NameError:
+        cmd = ['/usr/bin/rrdtool', 'update', fname, value]
+        proc = subprocess.Popen(cmd)
+        proc.communicate()  # wait until 'rrdtool update' completes
```
edquist, line 223: Does `rrdtool.update` return anything?  Because this rrd.update does not.

Also not sure if/how you want to handle non-success exit status from the rrdtool commandline.

templates/view_base.html
```diff
@@ -47,11 +47,7 @@
           <span class="icon-bar"></span>
           <span class="icon-bar"></span>
         </button>
-        {% if multice %}
-        <a href="/" class="navbar-brand">Grid Overview</a>
-        {% else %}
-        <a href="/" class="navbar-brand">HTCondor-CE Overview</a>
-        {% endif %}
+        <a href="/" class="navbar-brand">{{ self.title() }}</a>
```
edquist, line 50: so much nicer...

src/condor_ce_metric
```diff
@@ -44,10 +43,10 @@ def main():
     environ = {'htcondorce.spool': spooldir}
 
     fname = rrd.check_rrd(environ, None, "metrics", opts.group, opts.name)
-    rrdtool.update(fname, "N:%s" % str(value))
+    rrd.update(fname, "N:%s" % str(value))
```
matyasselmeci, line 46:
```suggestion
    rrd.update(fname, "N:%s" % value)
```
%s already converts to str

src/condor_ce_metric
```diff
@@ -44,10 +43,10 @@ def main():
...
 
     fname = rrd.check_rrd(environ, opts.spoof.split(":")[-1], "metrics", opts.group, opts.name)
-    rrdtool.update(fname, "N:%s" % str(value))
+    rrd.update(fname, "N:%s" % str(value))
```
matyasselmeci, line 49:
```suggestion
    rrd.update(fname, "N:%s" % value)
```

templates/vos.html
```diff
@@ -1,7 +1,14 @@
...
 
-{% block title %}HTCondor-CE VO Overview{% endblock %}
+{% block title %}
+{% if multice %}
+Grid Overview
+{% else %}
+HTCondor-CE Overview
+{% endif %}
+{% endblock %}
+
```
edquist, line 11: can we use that `{{ self.title() }}` thing here too?

src/htcondorce/rrd.py
matyasselmeci, line 9: nothing wrong with this, although ImportError continues to work in Python 3

matyasselmeci (approved):
A few comments but otherwise looks good; pre-approving

edquist (approved):
a few comments, nothing critical

i can't vouch at all for the jinja stuff btw

templates/vos.html
brianhlin: Nope, this is where we define the `title` for this page. If we wanted to refer to it later in either the base or child template, then we could use `self.title()` again

templates/view_base.html
brianhlin: Yeah, it's crazy what nuggets I find when I actually read the documentation :D

src/htcondorce/rrd.py
brianhlin: I had thought so too but I was seeing failures in the tests without this change

src/htcondorce/rrd.py
brianhlin: Not sure but we weren't checking for the return code before so /shrug

src/htcondorce/rrd.py
brianhlin: Oh actually I think maybe it was failing for another library. Ah well, this is more precise.

brianhlin: Yup, see https://github.com/htcondor/htcondor-ce/pull/406#discussion_r564889353

src/htcondorce/rrd.py
edquist: Ok, in that case if the rrdtool module is present they'll behave the same.  Seems that rrdtool.update raises exceptions on errors though.  If you want to do the same for the command line version, you could replace your Popen().communicate() with check_call() .

Or maybe you don't care, in which case who can blame you.

brianhlin: Yeah, that's a good suggestion so I'll drop in some `# TODO` comments but I'm not too worried about it for the time being

PR #407, opened 2021-01-26
Remove circular definition that caused condor_gangliad to thrash (stable)

brianhlin:
(HTCONDOR-161)


edquist (approved):
tbh i don't know enough about these definitions to comment competently, but if it fixes the problem, great

matyasselmeci (approved):

PR #410, opened 2021-02-03
Prefer GSI authN for CEs advertising to Central Collectors (HTCONDOR-237)

brianhlin:
We really want to prefer SSL but SSL creds are expected in /etc/pki/
locations. The grid world OTOH uses /etc/grid-security so HTCondor-CE
doesn't look in the right place to grab SSL creds and HTCondor happily
attempts SSL auth with no creds, failing the entire transaction.

https://opensciencegrid.atlassian.net/browse/HTCONDOR-237


matyasselmeci:
I think if GSI fails then it's not even going to go on to try SSL.

brianhlin:
> I think if GSI fails then it's not even going to go on to try SSL.

Yup, that's fine, since existing clients also prefer GSI (https://github.com/htcondor/htcondor-ce/blob/master/config/01-common-auth-defaults.conf#L19) and when we drop GSI in HTCondor-CE 6, new clients will be able to use SSL and we can document that host certs/keys need to be installed in the `/etc/pki` locations

Ticket to fail SSL on client failure to retrieve its own credentials: https://opensciencegrid.atlassian.net/browse/HTCONDOR-236

zmiller-uw:
A note to clarify: Unless something is broken (and it certainly has been in the past), if one authentication method fails, HTCondor should move on and try other methods.

One issue  that sometimes leads to problems is if the AUTHENTICATION actually succeeds, but the resulting entity is not AUTHORIZED.  In that case, it will not go back and try to authenticate with other methods.

-----Original Message-----

brianhlin:
@matyasselmeci given the above, are you ok with merging this?

matyasselmeci (approved):
Let's see how it goes.

PR #411, opened 2021-02-04
Ensure that the submitted code is an integer (HTCONDOR-241)

brianhlin:


src/htcondorce/registry.py
```diff
@@ -77,7 +77,14 @@ def osgid_to_ce(osgid):
 def fetch_tokens(reqid, config):
     binary = config.get('CONDOR_TOKEN_REQUEST_LIST', 'condor_token_request_list')
     pool = config.get('CONDORCE_COLLECTOR')
-    args = [binary, '-reqid', str(reqid), '-json']
+
+    try:
+        int(reqid)
+    except ValueError:
+        raise CondorToolException("Received invalid code: %s" % str(reqid))
```
matyasselmeci, line 84:
```suggestion
        raise CondorToolException("Received invalid code: %s" % reqid)
```

src/htcondorce/registry.py
```diff
@@ -100,7 +107,14 @@ def fetch_tokens(reqid, config):
 def approve_token(reqid, config):
     binary = config.get('CONDOR_TOKEN_REQUEST_APPROVE', 'condor_token_request_approve')
     pool = config.get('CONDORCE_COLLECTOR')
-    args = [binary, '-reqid', str(reqid)]
+
+    try:
+        int(reqid)
+    except ValueError:
+        raise CondorToolException("Received invalid code: %s" % str(reqid))
```
matyasselmeci, line 114:
```suggestion
        raise CondorToolException("Received invalid code: %s" % reqid)
```

matyasselmeci (commented):
str() is redundant with %s
Also, are negative numbers OK or do you want to catch those too?

brianhlin:
> Also, are negative numbers OK or do you want to catch those too?

Good call! Pushed a commit for that

src/htcondorce/registry.py
```diff
@@ -77,7 +77,14 @@ def osgid_to_ce(osgid):
...
     pool = config.get('CONDORCE_COLLECTOR')
-    args = [binary, '-reqid', str(reqid), '-json']
+
+    try:
+        assert int(reqid) > 0
+    except (ValueError, AssertionError):
+        raise CondorToolException("Received invalid code: %s" % reqid)
+    else:
+        args = [binary, '-reqid', str(reqid), '-json']
+
```
edquist, line 87: What is the explicit str conversion for?  (`str(reqid)`)  Is `reqid` sometimes input as an `int` object?  (I see it's always `request.form.get('code')`, which i would have guessed was already a `str`)

Anyway looks fine how you have it.  Might be tempting to tuck the try/except check into its own function, but that's up to you.

brianhlin: > What is the explicit str conversion for? (str(reqid)) Is reqid sometimes input as an int object? (I see it's always request.form.get('code'), which i would have guessed was already a str)

Yeah, I imagine you're correct but I don't think I want to change it for the time being. In the Python 3 version I'll add some type annotations

> Anyway looks fine how you have it. Might be tempting to tuck the try/except check into its own function, but that's up to you.

Done!

edquist (approved):

src/htcondorce/registry.py
```diff
@@ -74,10 +74,22 @@ def osgid_to_ce(osgid):
     return ces
 
 
+def validate_code(reqid):
+    """Ensure that the code we receive from the form submission is a positive integer
+    """
+    try:
+        assert int(reqid) > 0
+    except (ValueError, AssertionError):
```
matyasselmeci, line 82: I like this.

PR #412, opened 2021-02-04
Add support for whole node GPU jobs (HTCONDOR-103)

brianhlin:


brianhlin:
@matyasselmeci flightworthy is occupied with 9.0.0 this week so I'm reassigning the review to you. If it helps, I believe @juztas and some others in CMS are already running with this change.

src/condor_ce_router_defaults
```diff
@@ -68,6 +68,17 @@ JOB_ROUTER_DEFAULTS = """JOB_ROUTER_DEFAULTS_GENERATED @=jrd
...
 
+    /* Request GPUs for whole node jobs (HTCONDOR-103) */
+    copy_RequestGPUs = "orig_RequestGPUs";
+    eval_set_OriginalGPUs = ifThenElse(orig_RequestGPUs isnt undefined, orig_RequestGPUs, undefined);
+    set_GlideinGPUsIsGood = !isUndefined(MATCH_EXP_JOB_GLIDEIN_GPUs) && (int(MATCH_EXP_JOB_GLIDEIN_GPUs) isnt error);
+    set_JOB_GLIDEIN_GPUs = "$$(ifThenElse(WantWholeNode is true, !isUndefined(TotalGPUs) ? TotalGPUs : JobGPUs, OriginalGPUs))";
+    set_JobGPUs = JobIsRunning ? int(MATCH_EXP_JOB_GLIDEIN_GPUs) : OriginalGPUs;
+    set_RequestGPUs = ifThenElse((WantWholeNode is true && OriginalGPUs isnt undefined),
+                                 (!isUndefined(TotalGPUs) && TotalGPUs > 0)? TotalGPUs : JobGPUs,
+                                 OriginalGPUs);
```
matyasselmeci, line 79:
```suggestion
    /* This will also set MATCH_EXP_JOB_GLIDIEIN_GPUs when the job matches */
    set_JOB_GLIDEIN_GPUs = "$$(ifThenElse(MY.WantWholeNode is true, !isUndefined(TARGET.TotalGPUs) ? TARGET.TotalGPUs : JobGPUs, OriginalGPUs))";
    set_JobGPUs = MY.JobIsRunning ? int(MATCH_EXP_JOB_GLIDEIN_GPUs) : OriginalGPUs;
    set_RequestGPUs = ifThenElse((MY.WantWholeNode is true && OriginalGPUs isnt undefined),
                                 (!isUndefined(TARGET.TotalGPUs) && TARGET.TotalGPUs > 0) ? TARGET.TotalGPUs : JobGPUs,
                                 OriginalGPUs);
    set_GlideinGPUsIsGood = !isUndefined(MATCH_EXP_JOB_GLIDEIN_GPUs) && (int(MATCH_EXP_JOB_GLIDEIN_GPUs) isnt error);
```

A bit of tweaking for clarity. (I added some MYs and TARGETs, a comment, and moved set_GlideinGPUsIsGood to the end (because it depends on JOB_GLIDEIN_GPUs)).

src/condor_ce_router_defaults
```diff
@@ -68,6 +68,17 @@ JOB_ROUTER_DEFAULTS = """JOB_ROUTER_DEFAULTS_GENERATED @=jrd
                                                   default_queue,
                                                   "")));
 
+    /* Request GPUs for whole node jobs (HTCONDOR-103) */
+    copy_RequestGPUs = "orig_RequestGPUs";
+    eval_set_OriginalGPUs = ifThenElse(orig_RequestGPUs isnt undefined, orig_RequestGPUs, undefined);
```
matyasselmeci, line 73: You _might_ be able to simplify this to
```suggestion
    eval_set_OriginalGPUs = orig_RequestGPUs;
```
... along with a comment on why it's needed in the first place. Not certain though.

matyasselmeci (approved):
I think it does the job but could use a quick look from one of the HTCondor devs.

src/condor_ce_router_defaults
brianhlin: In my testing this seems to work!

src/condor_ce_router_defaults
brianhlin: I think this may actually break the intended behavior: outside of the match expression `MY` refers to the job route's attributes whereas `TARGET` refers to the pre-routed job's attributes. I'd need @johnkn to weigh in if `MY` vs `TARGET` within a match exp within a job route refers to the slot and pre-routed job ads, respectively.

johnkn: TARGET refers to the job being routed only when used in an EVALSET, so these changes look ok to me, other than the fact that they are confusing in the context of the job router route.   

However,  "moved set_GlideinGPUsIsGood to the end (because it depends on JOB_GLIDEIN_GPUs"  is potentially confusing because it implies that these SET operations will be applied in the order they are declared, which is not the case for job routes in the old job router syntax

brianhlin: @matyasselmeci given the above, could you submit a suggestion that only adds `MY` and `TARGET` in the match expression?

juztas: Do you need to change *_CPUs also then and add `MY` and `TARGET`? Or these are treated differently in HTCondor world (that would confuse me)

brianhlin: @juztas adding `MY` and `TARGET` are not strictly necessary. The only difference it could make in the `set_JOB_GLIDEIN_*` lines would be if the incoming jobs had a `Total*` attribute or the slot had a `WantWholeNode` attribute.

PR #413, opened 2021-02-05
Do not set HOME in the default job environment (HTCONDOR-176)

brianhlin:
This flips the switch here: https://github.com/htcondor/htcondor-ce/blob/V5-branch/src/condor_ce_router_defaults#L28-L30


matyasselmeci (approved):
does what it says it does

PR #416, opened 2021-02-10
python2 print remnant

thdesy:
clashing with future print/python3

### DIFF
--- a/src/condor_ce_info_status
+++ b/src/condor_ce_info_status
@@ -171,7 +171,7 @@ def main(argv):
 
     if opts.verbose:
         for res in resources:
-            print res
+            print( res )
     elif opts.submitfile:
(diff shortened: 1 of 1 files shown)


src/condor_ce_info_status
```diff
@@ -171,7 +171,7 @@ def main(argv):
 
     if opts.verbose:
         for res in resources:
-            print res
+            print( res )
```
brianhlin, line 174:
```suggestion
            print(res)
```

brianhlin (approved):
Minor styling suggestion but otherwise LGTM! FWIW, I do hope to tag HTCondor-CE 5.0.0 today, which has completely moved to Python 3! I'm aiming for it to go into the HTCondor development yum repositories.

PR #418, opened 2021-02-10
Combine v3, v4, and v5 docs into the most recent branch

brianhlin:
Instead, separate docs with nav tabs


brianhlin:
All link failures are due to htcondor-ce.org still pointing at readthedocs.io. Merging.

PR #425, opened 2021-02-11
Add mapfiles.d dirs to HTCondor-CE mapfile (HTCONDOR-244)

brianhlin:


config/condor_mapfile
```diff
@@ -1,31 +1,18 @@
...
-# callout mechanism (which may involve plugins such as LCMAPS or Argus).
-GSI (.*) GSS_ASSIST_GRIDMAP
-SSL "/CN=([-.A-Za-z0-9/= ]+)" \1@unmapped.htcondor.org
-CLAIMTOBE .* anonymous@claimtobe
-FS "^(root|condor)$" \1@daemon.htcondor.org
-FS "(.*)" \1
+###############################################################################
+
+# Place local authentication mapping customizations
+# in /etc/condor-ce/config.d
```
johnkn, line 14: I think this comment is wrong?

brianhlin:
@matyasselmeci can you take a look over the packaging changes?

config/condor_mapfile
brianhlin: Yup, fixed

matyasselmeci (approved):

brianhlin:
@matyasselmeci @johnkn requesting re-review since I added a commit to set `CERTIFICATE_MAPFILE_ASSUME_HASH_KEYS = True`, which changes the format of mapfiles so that regex in the second field have to be enclosed by `/`

matyasselmeci (approved):

PR #426, opened 2021-02-16
Use GlobalJobId as unique job identifier

drebatto:
Instead of making up a custom LRMS job identifier, use the Condor-defined GlobalJobId.
This doesn't introduce arbitrary convention on the JobId format.
It also allows using the field directly in condor_history constraints.

### DIFF
--- a/contrib/apelscripts/condor_batch.sh
+++ b/contrib/apelscripts/condor_batch.sh
@@ -32,8 +32,7 @@ safe_config_val SCALING_ATTR APEL_SCALING_ATTR
 safe_config_val BATCH_HOST APEL_BATCH_HOST
 
 TZ=GMT condor_history -constraint "$CONSTR" \
-    -format "%s" clusterId \
-    -format ".%s_${BATCH_HOST}|" ProcId \
+    -format "%s|" GlobalJobId \
     -format "%s|" Owner \
     -format "%d|" RemoteWallClockTime \
     -format "%d|" RemoteUserCpu \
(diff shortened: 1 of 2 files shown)


brianhlin:
Hi @drebatto thanks for the contribution! I'll have to admit that I'm not familiar with all of the APEL accounting rules. Are `#` characters valid in APEL records? Also, IIRC, the blah lrmsID needs to match the batch job ID to create the combined APEL record so I think you'd need to make a similar change to `condor_blah.sh`

maxfischer2781:
The same ID is generated in ``condor_blah.sh`` indeed:
https://github.com/htcondor/htcondor-ce/blob/1b3cec2b9012fec3e18c54d2bedac04a8c76a03f/contrib/apelscripts/condor_blah.sh#L42-L43

Both need to match, since [APEL joins based on these](https://github.com/apel/apel/blob/a94d44676df5863ae49bf9b4478e38541ad23a47/schemas/client.sql#L321-L324):
```
    INNER JOIN BlahdRecords ON BlahdRecords.ValidFrom <= EventRecords.EndTime AND
                             BlahdRecords.ValidUntil > EventRecords.EndTime AND
                             EventRecords.SiteID = BlahdRecords.SiteID AND EventRecords.JobName = BlahdRecords.LrmsId
                             AND SpecRecords.SiteID = BlahdRecords.SiteID AND SpecRecords.CEID = BlahdRecords.CEID
```

drebatto:
Yes, you are right. I modified `condor_blah.sh` too.
By the way, I implemented this patch because this is how we accounted for jobs with CREAM in the last six or seven years, and I wanted historical data to look consistent. For the same reason, I'll understand if you don't want to merge the patch and change the job id format for existing instances.

maxfischer2781:
I think this is a good change, even if it conflicts with the earlier format. The current format is not unique if a machine is re-installed, thereby resetting the ``clusterId`` counter. Since the ``GlobalJobId`` includes the QDate, it stays unique in this case.

brianhlin (approved):
LGTM, thanks!

PR #429, opened 2021-03-04
Add HTCondor-CE version information to condor_ce_version

brianhlin:
Currently, it only prints the HTCondor version, which is helpful but not quite complete.


timtheisen (approved):
LGTM

Issue #431, opened 2021-03-10
APEL - Make apelclient and ssmsend optional in condor_ce_apel.sh


drebatto:
Hi! 
The script `/usr/share/condor-ce/condor_ce_apel.sh` would need a little modification. The last two steps, namely `apelclient` and `ssmsend`, should be made optional because in many deployments APEL uses a central publisher, so the CEs don't send usage records to the central accounting system directly.
I'd propose to use an 'APEL_USE_PUBLISHER' configuration knob somewhere in the APEL-related configuration files to skip the last two steps.
I'll try to make a patch and create a pull request, but there are many ways to implement that so I hope to get the "CE-style" right.

brianhlin:
Fixed by #433

PR #432, opened 2021-03-10
Make apelclient and ssmsend optional in apel cron script

drebatto:
See issue #431

### DIFF
--- a/contrib/apelscripts/50-ce-apel-defaults.conf
+++ b/contrib/apelscripts/50-ce-apel-defaults.conf
@@ -23,6 +23,7 @@ APEL_OUTPUT_DIR = /var/lib/condor-ce/apel/
 # The attribute in the batch system that stores the scaling factor
 APEL_SCALING_ATTR = MachineAttrApelScaling0
 
-# If True, skip 'apelclient' and 'ssmsend' in the cron script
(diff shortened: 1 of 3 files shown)


brianhlin (commented):
Hi @drebatto, thanks for the PR! This looks ok to me but I'd like @edquist's eyes on the bash bits first. Also, I'd be interested in a different configuration name: maybe something like `APEL_SEND_RECORDS` where we use `apelclient` + `ssmsend` when set to `True` (the default). I'd prefer it this way since there's a direct relationship with enabling/disabling a specific HTCondor-CE feature whereas `APEL_USE_PUBLISHER` makes it seem like the CE offers a "publisher" feature.

drebatto:
Hi @brianhlin, it makes perfect sense. This is what I meant about getting the "CE-style" right :-)
I am going to make the change. By the way, I was wondering whether the bash function that wraps condor_config_val deserves a separate file, as it is used in three different scripts.

brianhlin:
Hi @drebatto thanks for making the changes! We're doing some reorganizing of these scripts for the next release of HTCondor-CE 5 so we'll handle the function duplication on our end. That reminds me, though: would you be able to rebase these changes on `V5-branch` and re-submit the PR against that branch?

brianhlin:
This has been resolved by #433, which resolves #431

PR #433, opened 2021-03-11
Make apelclient and ssmsend optional

drebatto:


brianhlin (approved):
LGTM and closes #432. Overriding failing CI for OSG EL8 since these failures are already occurring in `V5-branch`

PR #434, opened 2021-03-15
Require certs from CEs authenticating with SSL (HTCONDOR-344)

brianhlin:
This is intended for HTCondor-CE 5.1.0, which will ship with HTCondor 9.0.0


brianhlin:
Marking as `do-not-merge` until we have a specific version to reference for [this change](https://opensciencegrid.atlassian.net/browse/HTCONDOR-236)

johnkn (approved):

PR #437, opened 2021-03-19
Fixup walltime removal expression (HTCONDOR-350)

brianhlin:
It appears that JobCurrentStartExecutingDate does not get set in the
job ad so the expression nearly evaluates to True unless the admin
sets their max walltime to 50-some years


JaimeFrey (approved):

brianhlin:
Overriding unrelated EL8 CI failure

PR #441, opened 2021-03-25
Use PER_JOB_HISTORY_DIR for APEL reporting (HTCONDOR-293)

markcoatsworth:
This is my first pass at this work, I expect it will need some revisions. Please have a look and let me know.


contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -0,0 +1,76 @@
...
+
+safe_config_val () {
+    var=$1
+    attr=$2
+    val=$(condor_ce_config_val $attr) ||
+    fail "Failed to retrieve CE configuration value '$attr'"
+    eval "$var"='$val'
+}
+
+OUTPUT_DIR="$(condor_ce_config_val APEL_OUTPUT_DIR)"
```
brianhlin, line 16:
```suggestion
safe_config_val OUTPUT_DIR APEL_OUTPUT_DIR
```

contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -0,0 +1,76 @@
...
+[[ -d $OUTPUT_DIR && -w $OUTPUT_DIR ]] || fail "Cannot write to $OUTPUT_DIR"
+OUTPUT_DATETIME=`date '+%Y%m%d-%H%M'`
+
+HISTORY_DIR=`condor_config_val PER_JOB_HISTORY_DIR`
```
brianhlin, line 20: I forget exactly why @edquist had to implement `safe_config_val` but we probably should do the same thing for `condor_config_val`

contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -0,0 +1,76 @@
...
+QUARANTINE_DIR=$HISTORY_DIR/quarantine
+[[ -d $QUARANTINE_DIR && -w $QUARANTINE_DIR ]] || fail "Cannot write to $QUARANTINE_DIR"
+
+HISTORY_EXTRA_ARGS=(-format "\n" EMPTY)
```
brianhlin, line 24: Since we're not using `condor_history` anymore, we should probably rename the var to `CONDOR_Q_EXTRA_ARGS` or something

contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -0,0 +1,76 @@
...
+for file in $HISTORY_DIR/*
+do
+    # If $file is a directory, skip it
+    if [[ -d $file ]]; then
+        continue
+    fi
+
+    # Check if $file is a valid history file by looking for a ClusterId value
+    # If not found, assume the file is invalid and move to quarantine folder
+    if [[ -z "`condor_q -job $file -format \"%s\" ClusterId`" ]]; then
```
brianhlin, line 42:
```suggestion
    if [[ -z `condor_q -job $file -format "%s" ClusterId` ]]; then
```

You shouldn't need the `"` inside double brackets

contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -0,0 +1,76 @@
...
+HISTORY_EXTRA_ARGS=(-format "\n" EMPTY)
+safe_config_val SCALING_ATTR APEL_SCALING_ATTR
+[[ -z $SCALING_ATTR ]] || HISTORY_EXTRA_ARGS=(-format "%v|" "${SCALING_ATTR}" "${HISTORY_EXTRA_ARGS[@]}")
+
+safe_config_val BATCH_HOST APEL_BATCH_HOST
+safe_config_val CE_HOST APEL_CE_HOST
+safe_config_val CE_ID APEL_CE_ID
+
+# Iterate over all the files in PER_JOB_HISTORY_DIR
+for file in $HISTORY_DIR/*
```
brianhlin, line 33:
```suggestion
for file in "$HISTORY_DIR/*"
```

We should wrap this var in double-quotes for safety since we get its value from an external source

contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -0,0 +1,76 @@
...
+do
+    # If $file is a directory, skip it
+    if [[ -d $file ]]; then
```
brianhlin, line 36:
```suggestion
    if [[ ! -f $file ]]; then
```

We probably want to just skip everything that's not a regular file

brianhlin (changes requested):
A few minor changes but overall LGTM. I'm on board with using a missing `ClusterID` as a proxy for a bad history file.

contrib/apelscripts/condor_batch_blah.sh
matyasselmeci:
```suggestion
for file in "$HISTORY_DIR"/*
```
You don't want to quote the `*` -- it would be taken literally.

contrib/apelscripts/condor_batch_blah.sh
edquist: > I forget exactly why @edquist had to implement `safe_config_val` but we probably should do the same thing for `condor_config_val`

Heard my name!

Looks like the idea behind `safe_config_val` was to provide a single function to get a config val if it exists, otherwise cause the script fail.

@brianhlin, it looks like you introduced `safe_config_val` originally [here](https://github.com/htcondor/htcondor-ce/commit/d8277b21508c96ec09782fbb0204b68b63e9dee0#diff-de4e3077e8b9dd3874333d5c4124d541d39f719fd42a5a99bd02e8567aec0c0eR3-R5)

but i modified it later (in 8a8c052992a163306bd204b410a2e6c7212c6fa7) because the `exit 1` was happening inside a subshell and not getting propagated to the main script process

markcoatsworth: @brianhlin So I'm assuming I should make a second function called `safe_ce_config_val()` and change `safe_config_val()` to call `condor_config_val`?

brianhlin: Yeah, that sounds reasonable

PR #442, opened 2021-03-25
Allow CE clients to use SSL to query the schedd/collector

brianhlin:
(HTCONDOR-366)


bbockelm (approved):
LGTM.

As a follow-up, it seems reasonable to also enable this for the other authorization levels.

brianhlin:
Cowardly only adding it for SchedD/Collector WRITE to allow CE admins to add SSL-based mappings

brianhlin:
Ignoring unrelated EL8 osg-test failure

PR #447, opened 2021-03-30
Disable EL8 builds; add RPM build validation to PR CI

brianhlin:
4.5.0 RPM builds failed due to EL8, which we don't support in v4 anyway

### DIFF
--- a/.github/workflows/build_rpms.yaml
+++ b/.github/workflows/build_rpms.yaml
@@ -1,0 +1,33 @@
+name: Test building of HTCondor-CE RPMs
+on: [pull_request]
+
+jobs:
+  build-rpms-and-upload:
+    runs-on: ubuntu-latest
+    if: startsWith(github.repository, 'htcondor/')
+    strategy:
+      matrix:
+        dver: [7]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Prepare Docker
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          docker run --detach --env "container=docker" \
+                 --name $GITHUB_SHA \
+                 --volume `pwd`:/htcondor-ce:rw \
+                 centos:centos${{ matrix.dver }} \
+                 /usr/sbin/init
+
+      - name: Build CHTC EL${{ matrix.dver }} RPMs
+        run: |
+          docker exec $GITHUB_SHA \
```
edquist, line 29: Sorry where does this `$GITHUB_SHA` come from?

edquist: It seems a bit confusing to me to use the git commit sha as the docker image name - like, "wait is that github sha a docker image hash, or just a name?  It's a name but it's not obvious when you look at it.
But i guess if that's the model you want then it works.

Even tacking on a prefix or suffix to the name could make it a bit more clear.  Eg, `auto-build.$GITHUB_SHA` - or some more descriptive name.

edquist (approved):
Probably fine as is, but some food for thought in my comments

PR #448, opened 2021-03-30
fix syntax issues with new job router transforms for HTCONDOR-243

johnkn:
$() Expansions my not have internal spaces
  evals happening at the wrong time for CERequirements and OnExitHold


brianhlin (approved):
LGTM, thanks!

PR #451, opened 2021-04-01
Update condor_ce_startup

xzhao87:
For sites where the user 'condor' has a primary group not named 'condor', the verification will fail, therefore prevents condor-ce from starting. The change is to loose it, only verify the required dirs are owned by 'condor' user, regardless of the group name.

### DIFF
--- a/src/condor_ce_startup
+++ b/src/condor_ce_startup
@@ -14,7 +14,7 @@ fail() {
 # Verify that required dirs are owned by condor:condor (SOFTWARE-2806)
 required_dirs="/var/run/condor-ce /var/log/condor-ce /var/log/condor-ce/user /var/lib/condor-ce /var/lib/condor-ce/spool
 /var/lib/condor-ce/execute /var/lock/condor-ce /var/lock/condor-ce/user"
-bad_dirs=$(stat --format "%n (%U:%G)" $required_dirs | awk '$2 != "(condor:condor)" {print $0}')
+bad_dirs=$(stat --format "%n %U" $required_dirs | awk '$2 != "condor" {print $0}')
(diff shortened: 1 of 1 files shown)


src/condor_ce_startup
```diff
@@ -14,7 +14,7 @@ fail() {
 # Verify that required dirs are owned by condor:condor (SOFTWARE-2806)
 required_dirs="/var/run/condor-ce /var/log/condor-ce /var/log/condor-ce/user /var/lib/condor-ce /var/lib/condor-ce/spool
 /var/lib/condor-ce/execute /var/lock/condor-ce /var/lock/condor-ce/user"
-bad_dirs=$(stat --format "%n (%U:%G)" $required_dirs | awk '$2 != "(condor:condor)" {print $0}')
+bad_dirs=$(stat --format "%n %U" $required_dirs | awk '$2 != "condor" {print $0}')
 [ -z "$bad_dirs" ] || fail 1 "The following directories are not owned by the condor user/group:\n$bad_dirs"
```
brianhlin, line 18:
```suggestion
[ -z "$bad_dirs" ] || fail 1 "The following directories are not owned by the condor user:\n$bad_dirs"
```

brianhlin (changes requested):
Hi @xzhao87, thanks for submitting the PR! I've made one suggestion to update the failure message but once that's fixed this LGTM. And FWIW, the CI failures are due to a known issue in the tests so they're not related

brianhlin (approved):
LGTM. Thanks again!

brianhlin:
Looking at the CI failures, `osg-test` failures are a known issue and the CE starts up fine. Merging.

PR #452, opened 2021-04-02
Update HTCondor-CE registry to Python 3 (HTCONDOR-307)

markcoatsworth:
First pass at updating the registry to python3, removing Genshi and replacing it with Jinja2. This work is based on BrianL's previous work doing the same with the CE view scripts. Likely some things missing here.

### DIFF
--- a/src/htcondorce/registry.py
+++ b/src/htcondorce/registry.py
@@ -10,11 +10,11 @@ from urllib.request import urlopen
 from urllib.error import URLError
 
 from flask import Flask, Config, url_for, make_response, request
-import genshi.template
+from jinja2 import Environment, FileSystemLoader, select_autoescape
 import subprocess
 from tools import to_str
 
-_loader = None
+_jinja_env = None
 
 TOPOLOGY_RG = "https://topology.opensciencegrid.org/rgsummary/xml"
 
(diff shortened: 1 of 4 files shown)


src/htcondorce/registry.py
```diff
@@ -139,83 +139,85 @@ def create_app(test_config = None):
     else:
         app.config.update(test_config)
 
-    global _loader
+    global _jinja_env
     template_location = app.config.get('HTCONDORCE_TEMPLATES', '/usr/share/condor-ce/templates')
-    _loader = genshi.template.TemplateLoader(template_location, auto_reload=True)
+    _jinja_env = Environment(loader=FileSystemLoader(template_location),
+                             autoescape=select_autoescape(['html', 'xml']),
+                             auto_reload=True)
```
brianhlin, line 146:
```suggestion
                             autoescape=select_autoescape(['html', 'xml']))
```

I don't think we need to enable auto reload since we're loading static files on disk

brianhlin (approved):
One suggestion but otherwise LGTM!

src/htcondorce/registry.py
markcoatsworth: I set this to auto reload because that's the way it was working under Genshi. If you think we can disable it, that's fine with me.

brianhlin: Ah good point. I don't think it's strictly necessary but I'm not interested in experimenting to find out here!

brianhlin:
Ci failures expected, merging

PR #453, opened 2021-04-05
Add accounting group user maps (HTCONDOR-187)

brianhlin:
Includes #450 but moves the config into the local batch condor specific config files, changes the map names, adds example map files, and always sets `JOB_ROUTER_CLASSAD_USER_MAP_NAMES`


brianhlin:
The CI failure here seems to actually uncover a problem with HTCondor where the client segfaults on EL8 when using tokens. This, however, is most likely an underlying HTCondor issue.

johnkn (commented):
Looks fine to me

PR #455, opened 2021-04-07
Update packaging for 5.1.0; drop (noreplace) for condor_mapfile

brianhlin:
We can do the latter since we will only ship 5.x along with 9.0.x in the CHTC and OSG 3.5 upcoming/3.6 Yum repos

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ project(condor-ce NONE)
 
 cmake_minimum_required(VERSION 2.6)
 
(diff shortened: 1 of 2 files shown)


matyasselmeci:
Do we have / can we add a comment in the condor_mapfile we ship to tell people where they should now add changes?

brianhlin:
Yup! https://github.com/htcondor/htcondor-ce/blob/V5-branch/config/condor_mapfile#L8-L9

matyasselmeci:
Nice! Approved.

brianhlin:
EL8 CI failure due to suspected SciTokens + condor_ping segfault. Unrelated to this change so I'm merging

PR #459, opened 2021-04-08
Fix check of job route names in CE configuration. HTCONDOR-406

JaimeFrey:
Split the value of JOB_ROUTER_ROUTE_NAMES on space/comma instead of
splitting into individual characters.


brianhlin (approved):
LGTM!

PR #460, opened 2021-04-08
Add Python linting CI

### LINKED ISSUE #457: This should be as exc and not as e
https://github.com/htcondor/htcondor-ce/blob/a8958aa7d5ec83fa5603ff1ce14c7400a40e7d7e/src/condor_ce_trace#L225

### LINKED ISSUE #458: network_iface.name is returned as bytes and errors on python3 should probably be converted to string
https://github.com/htcondor/htcondor-ce/blob/a8958aa7d5ec83fa5603ff1ce14c7400a40e7d7e/src/condor_ce_host_network_check#L238
str(network_iface.name)

brianhlin:
Not sure how folks will feel about `from htcondor import htcondor` and `from classad improt classad` but it makes pylint happy and seems to work ok. This also drops a bunch of unused variables and imports; fixes #457 and fixes #458


matyasselmeci:
`from htcondor import htcondor` gives me an ImportError when I try it by hand.

brianhlin:
Curious. The tests are starting to roll in and it looks like they're ok and it works fine for me manually:

```
[blin@lhcb-ce ~]$ python3
Python 3.6.8 (default, Nov 16 2020, 16:55:22) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from htcondor import htcondor
>>> from classad import classad
>>> htcondor.param['SEC_CLIENT_AUTHENTICATION_METHODS']
'CLAIMTOBE FS'
```

Are you using the python 2 bindings?

matyasselmeci:
Nope. Python 3 (on Fedora though so condor's 8.8.10)

matyasselmeci:
It _does_ work with htcondor-8.9.13 installed from pip.

brianhlin:
Yeah and `lhcb-ce` has `python3-condor = 8.9.13`. This also works on a CE with 8.9.11 so this appears to work after 8.9.10. I was thinking of adding an htcondor >= 9.0.0 version requirement to the packaging anyway because 8.9.12/8.9.13 are known to be broken for HTCondor-CE. I could also revert the change and have pylint ignore `E1101` across the board. Thoughts?

matyasselmeci:
```
try:
    from htcondor import htcondor
except ImportError:
    try:
        import htcondor
    except ImportError:
        print("you should install condor, buddy", file=sys.stderr)
```
perhaps?
With a comment saying that this makes the CI happy and real users don't need to do it?

matyasselmeci:
Another option is
```
import htcondor  # pylint: disable=E1101
```
for those specific imports?

brianhlin:
I've reverted the changes to the classad/htcondor imports in favor of just ignoring the pylint errors to avoid having to add an HTCondor version requirement

edquist (approved):
lgtm

(hopefully it works, right?)

brianhlin:
Overriding unrelated CI failure

PR #461, opened 2021-04-08
Merge JR entry name check from V4-branch (HTCONDOR-406)

brianhlin:
Let's see if I did this right...

### DIFF
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -85,7 +85,7 @@ def main():
             # The job router can function this way but it's likely a config error
             route_order = htcondor.param.get('JOB_ROUTER_ROUTE_NAMES', '')
             if route_order:
-                missing_route_def = set(route_order).difference(set(parse_route_names(config_val)))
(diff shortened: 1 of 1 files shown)


JaimeFrey:
I scanned the other commits from V4-branch that come along for the ride. None result in changes in V5-branch, though perhaps one should: 1bc652ed95f.
It changes a `pwd` to "$PWD" in .github/workflows/build_rpms.yml to avoid creating an extra process. The equivalent code in build_and_test_rpms.yml in V5-branch has `pwd`.

JaimeFrey (approved):
LGTM!

brianhlin:
Good catch @JaimeFrey! Do you think I should submit a separate PR with that commit included in the merge commit to maintain a reasonable `git merge-base`?

JaimeFrey:
If you merge this PR, then the pwd commit will be merged as part of it. A separate PR of that commit won't do anything. And it looks like the relevant bit of code was added in different commits on the two branches, so git doesn't know they're related. A new commit on V5-branch (possibly in this PR) with the same change seems like the best option.

JaimeFrey (approved):

PR #463, opened 2021-04-09
Fix default transform routes (HTCONDOR-243)

brianhlin:
A 'JOB_ROUTER_ROUTE_' will override a route in JOB_ROUTER_ENTRIES if
they have the same name, even if
'USE_JOB_ROUTER_DEPRECATED_ROUTER_ENTRIES = True' and the
'JOB_ROUTER_ROUTE_' is uses the transform syntax.

To avoid unexpectedly overrriding an existing CE's route (many of them
probably just stick with the default names we give), use a different
name in the transform syntax.

Also add some upgrade notes about the new transform syntax, including a
warning about transform routes that we provide by default. More detailed
docs about the transform syntax to come in another commit.

### DIFF
--- a/config/02-ce-bosco-defaults.conf
+++ b/config/02-ce-bosco-defaults.conf
@@ -18,7 +18,7 @@ JOB_ROUTER_ENTRIES = \
    ]
 
 
-JOB_ROUTER_ROUTE_Local_BOSCO @=jrt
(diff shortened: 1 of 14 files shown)


johnkn (approved):
Now that the dead routes in the *-defaults.conf files have been removed, and better names given for the tranform routes.  I approve.

PR #465, opened 2021-04-12
Downgrade CE config verify errors

brianhlin:
Now that we support job transforms, it's feasible that an admin won't
have JOB_ROUTER_ENTRIES or JOB_ROUTER_DEFAULTS defined (the latter
being more difficult since we generate them automatically). For now,
just downgrade this to a warning.

Doc URL doesn't exist yet, I've got a separate branch for it. See https://brianhlin.github.io/htcondor-ce/v5/configuration/writing-job-routes/#setting-job-environments for what I plan for it to look like

### DIFF
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -122,8 +122,8 @@ def main():
 
         # Ensure that users don't set the job environment in the Job Router
         if is_osg and any(x.endswith('environment') for x in entry.keys()):
-            error("Do not use the Job Router to set the environment. Place variables under "
-                  + "[Local Settings] in /etc/osg/config.d/40-localsettings.ini")
+            error("Do not use the Job Router to set the environment. See documentation for more details: "
+                  + "https://htcondor.github.io/htcondor-ce/v5/configuration/writing-job-routes/#setting-job-environments")
 
         # Warn users about eval_set_ default attributes in the ENTRIES since their
         # evaluation may occur after the eval_set_ expressions containg them in the
--- a/src/verify_ce_config.py
+++ b/src/verify_ce_config.py
@@ -8,6 +8,8 @@ import os
 import re
 import sys
 
+from collections import defaultdict
+
 
 def error(msg):
     """Exit with error 'msg'
@@ -19,6 +21,12 @@ def warn(msg):
     """Print warning 'msg'
     """
     print("WARNING: " + msg)
+
+
+def debug(msg):
+    """Print debug 'msg'
+    """
+    print("DEBUG: " + msg)
 
 
 def parse_route_names(entries_config):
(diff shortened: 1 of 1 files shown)


matyasselmeci:
Is there any way to detect if they _are_ using `JOB_ROUTER_ROUTE_*` transforms (or at least have them)? (Are those the 'new' way? I don't like the idea of admins getting warnings forever unless they add some dummy backward compat config or something.)

brianhlin:
Yes but I think the changes are too involved for 5.1.0 so I'll make a ticket for that and target it for 5.1.1 or 5.2.0

src/verify_ce_config.py
```diff
@@ -59,19 +59,22 @@ def main():
...
         except KeyError:
-            error("Missing required %s configuration value" % attr)
+            warn("Missing %s configuration value. " % attr +
+                 "Can be safely ignored if using 'JOB_ROUTER_ROUTE_*' ClassAd transforms.")
 
         # store the ads (iterating through ClassAdStringIterator consumes them)
         try:
             parsed_jr_ads[attr] = list(classad.parseAds(config_val))
         except ValueError:
             # We shouldn't ever get here since classad.parseAds() only raises ValueError when it's given
```
matyasselmeci, line 69: "We shouldn't ever get here" implies at least an error, if not an assertion.

src/verify_ce_config.py
```diff
@@ -59,19 +59,22 @@ def main():
...
             # non-string/non-file output and htcondor.param shouldn't contain such values
-            error("Failed to parse %s configuration value" % attr)
+            warn("Failed to parse %s configuration value. " % attr +
+                 "Can be safely ignored if using 'JOB_ROUTER_ROUTE_*' ClassAd transforms.")
 
         # If JRD or JRE can't be parsed, the job router can't function
         if not parsed_jr_ads[attr]:
-            error("Could not read %s in the HTCondor-CE configuration." % attr)
+            warn("Could not read %s in the HTCondor-CE configuration. " % attr +
```
matyasselmeci, line 76: Is the admin going to get multiple warnings for the same thing?

src/verify_ce_config.py
brianhlin: Agreed but I'm not keen on keeping this an error that could prevent CE startup without knowing more about how an unset `JOB_ROUTER_ENTRIES` interacts with this piece of code.

src/verify_ce_config.py
brianhlin: Yes, once each for an unparseable JRE and JRD. The admin will really have to go out of their way to unset the latter, though.

src/verify_ce_config.py
matyasselmeci: Can you update the comment then?

matyasselmeci: Actually, now that KeyError doesn't exit (?), is it still true that we shouldn't get to the ValueError exception?
Or, worse, can
```
            parsed_jr_ads[attr] = list(classad.parseAds(config_val))
```
crash?

brianhlin: D'oh, good point

matyasselmeci:
It kinda feels like we're warning people for doing the right thing. We could also turn the warnings into debug messages until we get JOB_ROUTER_ROUTE detection done -- I'll encourage us to hurry with it :-)

The code looks overall fine except you should make sure that changes to the code flow caused by not bailing early won't cause exceptions at later points.

brianhlin:
@matyasselmeci updated the commit to address your concerns

matyasselmeci (approved):

PR #467, opened 2021-04-12
Fix IPv6 format

brianhlin:
Just `::1` was confusing gunicorn so it couldn't find the port and we listen on all addresses in IPv4 so we should do the same for IPv6


edquist (approved):
tbh i have no idea how `[::]1:<PORT>` works (vs `::1:<PORT>`).

But if it works for you, more power to it.

PR #468, opened 2021-04-13
Various router transform fixups (HTCONDOR-243)

brianhlin:
I haven't simplified the environment transform since the `{advertise_pilots}` string is shared between the transform and the `JOB_ROUTER_DEFAULTS`. I don't hardcode the string based on the various `DISABLE_PILOT_ADS` settings because I'd like users to be able to change that without having to restart the CE to see its effect.


johnkn (approved):
looks fine.

PR #471, opened 2021-04-14
Split up troubleshooting doc

brianhlin:
ToddT gave the go ahead to merge docs immediately with review after the fact


brianhlin:
Many broken links but that's fine right now since I just want to get the initial doc skeleton in place

PR #473, opened 2021-04-14
Add scitokens mapping example

brianhlin:


brianhlin:
Install failures for runs without HTCondor 9.0.0 available (new RPM requirement), EL8 OSG `condor_ce_ping` still aborting so it looks like we have a new HTCondor bug. Config no-op, builds work, merging.

PR #477, opened 2021-04-26
(HTCONDOR-420)  Unblock all signals at condor_ce_view start-up.

Todd-L-Miller:
It seems like "daemon core" daemons are expected to handle setting their own signal mask(s) on start-up.

### DIFF
--- a/src/condor_ce_view
+++ b/src/condor_ce_view
@@ -8,6 +8,15 @@ import socket
 import time
 import optparse
 
+# It looks like HTCondor spawns "daemon core" processes with a signal
+# mask set, probably because it doesn't want to get signals between the
+# fork() (clone()?) and exec().  These two lines of code unblock all
+# signals.  This doesn't cause this program to respond as you'd gunicorn
+# to, but they may be a problem with this script, instead.  (It would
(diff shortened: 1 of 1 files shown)


Todd-L-Miller:
For PATh metrics, please add a comment that begins with "code review" to HTCONDoR-420 when you're done here.

src/condor_ce_view
```diff
@@ -8,6 +8,15 @@ import socket
 import time
 import optparse
 
+# It looks like HTCondor spawns "daemon core" processes with a signal
+# mask set, probably because it doesn't want to get signals between the
+# fork() (clone()?) and exec().  These two lines of code unblock all
+# signals.  This doesn't cause this program to respond as you'd gunicorn
+# to, but they may be a problem with this script, instead.  (It would
+# have to forward signals along to the 'server' process, at the least.)
```
brianhlin, line 16:
```suggestion
# signals.  This doesn't cause this program to respond as you'd expect
# gunicorn to, but they may be a problem with this script, instead.  (It
# would have to forward signals along to the 'server' process, at the
# least.)
```

Missed a word

brianhlin (approved):
One suggestion but otherwise LGTM

PR #485, opened 2021-05-21
Change /var/lock/condor-ce/user to be 1777 to match the spec file

matyasselmeci:
https://opensciencegrid.atlassian.net/browse/HTCONDOR-512


brianhlin (approved):
This change looks good but it doesn't appear to solve our OSG VMU test woes

Issue #419, opened 2021-02-11
Jobs missing in APEL records versus condor_history


maxfischer2781:
We occasionally see a significant number of jobs missing from APEL – about %5 to 10%. The jobs are in `condor_history` but do not get written out to the APEL records:

```bash
$ condor_history -match 1 3100000.0 -file /backup/htcondor-ce-1-kit/condor_history_20210131
 ID     OWNER          SUBMITTED   RUN_TIME     ST COMPLETED   CMD
3100000.0   plhcb49         1/31 06:35   0+02:03:37 C   1/31 08:39 DIRAC__sCOtw_pilotwrapper.py
$ grep ^309999 /var/lib/condor-ce/apel/batch-202101*-htcondor-ce-1-kit  # jobs before
/var/lib/condor-ce/apel/batch-20210131-htcondor-ce-1-kit:3099992.0_htcondor-ce-1-kit.gridka.de|alicesgm|47738|42537|595|1612071724|1612119462|2046508|6599748|1|1.069035087719298|
/var/lib/condor-ce/apel/batch-20210131-htcondor-ce-1-kit:3099996.0_htcondor-ce-1-kit.gridka.de|atlasprd|23975|177846|5667|1612076854|1612100829|8745192|42251352|8|0.9114035087719299|
/var/lib/condor-ce/apel/batch-20210131-htcondor-ce-1-kit:3099991.0_htcondor-ce-1-kit.gridka.de|atlasprd|22693|169983|9477|1612076855|1612099548|9823420|43896844|8|0.9151754385964912|
/var/lib/condor-ce/apel/batch-20210131-htcondor-ce-1-kit:3099997.0_htcondor-ce-1-kit.gridka.de|atlasprd|20630|154714|10286|1612077309|1612097939|9061048|43332508|8|0.9151754385964912|
/var/lib/condor-ce/apel/batch-20210131-htcondor-ce-1-kit:3099993.0_htcondor-ce-1-kit.gridka.de|atlasprd|19886|146123|5000|1612076854|1612096740|8949180|42814016|8|0.9114035087719299|
/var/lib/condor-ce/apel/batch-20210131-htcondor-ce-1-kit:3099990.0_htcondor-ce-1-kit.gridka.de|alicesgm|19538|7415|364|1612071725|1612091263|2143792|9827460|1|0.9114035087719299|
[...]

PR #488, opened 2021-06-01
Safe defaults for APEL

maxfischer2781:
This PR adds defaults for APEL to ensure accounting does not skip jobs for a fresh installation.

* If a job's host did not define the ``APEL_SCALING_ATTR``, it is quarantined instead of reporting ``undefined`` (see https://github.com/apel/apel/issues/210)
* The default ``PER_JOB_HISTORY_DIR`` is created for user ``condor:condor`` (see #419)

The first change could alternatively set a default (as in an earlier commit). Since this may hide veritable configuration errors, excluding broken jobs and allowing manual recovery seemed preferable.

### DIFF
--- a/contrib/apelscripts/condor_batch_blah.sh
+++ b/contrib/apelscripts/condor_batch_blah.sh
@@ -32,7 +32,7 @@ mkdir -p "$QUARANTINE_DIR"
 
 CONDOR_Q_EXTRA_ARGS=(-format "\n" EMPTY)
 safe_ce_config_val SCALING_ATTR APEL_SCALING_ATTR
-[[ -z $SCALING_ATTR ]] || CONDOR_Q_EXTRA_ARGS=(-format "%v|" "${SCALING_ATTR} isnt undefined ? ${SCALING_ATTR} : 1" "${CONDOR_Q_EXTRA_ARGS[@]}")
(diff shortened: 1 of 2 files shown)


markcoatsworth (approved):
I could nitpick about using `! -z $SCALING_ATTR` for consistency, but this works just fine.

rpm/htcondor-ce.spec
```diff
@@ -360,6 +361,7 @@ fi
 %{_sysconfdir}/condor/config.d/50-condor-apel.conf
 %config(noreplace) %{_sysconfdir}/condor-ce/config.d/50-ce-apel.conf
 %attr(-,root,root) %dir %{_localstatedir}/lib/condor-ce/apel/
+%attr(-,condor,condor) %dir %{_localstatedir}/lib/condor/history/
```
markcoatsworth, line 364: Verified this setup in `lhcb-ce.chtc.wisc.edu`, looks good to me.

contrib/apelscripts/condor_batch_blah.sh
```diff
@@ -52,6 +52,12 @@ do
         mv "$file" "$QUARANTINE_DIR"
         continue
     fi
+    # Check that APEL scaling information is available if requested
+    # Otherwise, quarantine the job information to allow manual fixing
+    if [[ -n $SCALING_ATTR && -z `condor_q -job $file -format "%f" "${SCALING_ATTR}"` ]]; then
```
markcoatsworth, line 57: I could nitpick here about using `! -z $SCALING_ATTR` for consistency with previous checks, but this works just fine.

markcoatsworth (approved):
All looks good to me.

PR #490, opened 2021-06-03
Address auth doc  review comments (HTCONDOR-419)

brianhlin:
@zmiller I didn't address your point about `/etc/grid-security/certificates/` because it's touched on in this sentence:

> But traditionally, CEs and their clients have authenticated with each other using specialized grid certificates (e.g.
certificates issued by [IGTF CAs](https://dl.igtf.net/distribution/igtf/current/accredited/accredited.in)) located in `/etc/grid-security/`.

### DIFF
--- a/docs/v5/configuration/authentication.md
+++ b/docs/v5/configuration/authentication.md
@@ -14,7 +14,7 @@ Built-in Mapfiles
 ----------------
 
 HTCondor-CE uses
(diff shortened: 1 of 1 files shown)


brianhlin:
Zach approved this on the Jira ticket and the link failures are unrelated. Merging.

PR #492, opened 2021-06-15
Updated CE config verification to include JR transforms (HTCONDOR-527)

markcoatsworth:


markcoatsworth:
Abandoning this PR, will try again with a different approach.

PR #497, opened 2021-07-01
Updated CE verification script to handle job router transforms (HTCONDOR-527)

markcoatsworth:


src/verify_ce_config.py
```diff
@@ -103,35 +108,45 @@ def main():
...
-            warn("%s in JOB_ROUTER_ENTRIES " % ', '.join(no_effect_attr)
-                 + "may not have any effect. Use the 'set_' prefix instead.")
+    # Also warn about routes specified in JOB_ROUTER_ROUTE_<name> transforms that don't appear in the parsed JRE.
+    if jr_transforms:
+        for transform_attr, transform_value in jr_transforms:
+            missing_route_def = set(transform_value.replace(',',' ').split()).difference(set(parse_route_names(config_val)))
+            if missing_route_def:
+                warn("The following are specified in %s "
+                        "but cannot be found in JOB_ROUTER_ENTRIES: %s"
+                        % transform_attr, ', '.join(missing_route_def))
```
brianhlin, line 118: This isn't quite the right check to make (transforms are separate from entries). We should warn if there are route names showing up in `JOB_ROUTER_ROUTE_NAMES` that don't have a corresponding `JOB_ROUTER_ROUTE_<NAME>` config variable

Also, the usage of `config_val` will likely lead to confusion down the line since it's defined in the for loop and it luckily works here since the last iteration in the loop will set `config_val` the ads defined by `JOB_ROUTER_ENTRIES`

src/verify_ce_config.py
```diff
@@ -63,27 +66,29 @@ def main():
...
+            try:
+                parsed_jr_ads[attr] = list(classad.parseAds(config_val))
+            except ValueError:
+                # We shouldn't ever get here since classad.parseAds() only raises ValueError when it's given
+                # non-string/non-file output and htcondor.param shouldn't contain such values
+                debug("Failed to parse %s configuration value. " % attr)
+
+                # If JRD or JRE can't be parsed, the job router can't function
+                if not parsed_jr_ads[attr]:
+                    debug("Could not read %s in the HTCondor-CE configuration. " % attr)
```
brianhlin, line 91: I think we should revert this bit

brianhlin (changes requested):
Hey @markcoatsworth , a few comments but overall, I think we should revisit that refactoring idea that you had and just simplify it. For everyone's sanity, it makes sense to treat the new transform config and old classad config as mutually exclusive, i.e. if `JOB_ROUTER_ROUTE_*` is defined, only do JRT validations.

So I propose that we:

1.  Only run all of the old checks `if not jr_transform`
2.  If they have any JRTs defined, warn about names in `JOB_ROUTER_ROUTE_NAMES` that don't have a corresponding JRT
3. Since `main()` already has too many branches, split all of the old checks into a separate function

markcoatsworth:
My most recent commit (https://github.com/htcondor/htcondor-ce/pull/497/commits/4d9241d0e933692eba573f0bedd1a71c012752ce) addresses all these comments but it's also not showing up in this window? Let me know if there's anything I need to do on my end.

src/verify_ce_config.py
```diff
@@ -63,75 +66,85 @@ def main():
...
-    # Find all eval_set_ attributes in the JOB_ROUTER_DEFAULTS
-    eval_set_defaults = set([x.lstrip('eval_') for x in parsed_jr_ads['JOB_ROUTER_DEFAULTS'][0].keys()
-                             if x.startswith('eval_set_')])
-
-    # Find all default_ attributes used in expressions in the JOB_ROUTER_DEFAULTS
-    default_attr = set([re.sub(r'.*(default_\w*).*', 'eval_set_\\1', str(x))
+    # If JOB_ROUTER_ROUTE_<name> rules are defined, verify that these transforms exist
+    if jr_transforms:
+        route_names = htcondor.param.get("JOB_ROUTER_ROUTE_NAMES", "")
+        for name in route_names.split(" "):
```
brianhlin, line 74:
```suggestion
        for name in re.split(r'[, \t]+', route_names):
```

`JOB_ROUTER_ROUTE_NAMES` can be comma separated

src/verify_ce_config.py
```diff
@@ -63,75 +66,85 @@ def main():
...
+            route_def = htcondor.param.get(f"JOB_ROUTER_ROUTE_{name}", "")
+            if not route_def:
+                warn(f"The route {name} is specified in JOB_ROUTER_ROUTE_NAMES, but no corresponding JOB_ROUTER_ROUTE_{name} exists")
```
brianhlin, line 77: You could probably do this with a `set.difference()` but the list of routes are probably not large enough to matter

brianhlin (approved):
LGTM except some minor comments

src/verify_ce_config.py
```diff
@@ -63,27 +66,29 @@ def main():
 
     # Create dict whose values are lists of ads specified in the relevant JOB_ROUTER_* variables
     parsed_jr_ads = defaultdict(list)
+    jr_transforms = {key : value for key, value in htcondor.param.items() if "JOB_ROUTER_ROUTE_" in key}
```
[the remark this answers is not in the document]
brianhlin: Just tested this, it looks like the keys in `htcondor.param` are case insensitive

src/verify_ce_config.py
markcoatsworth: Let's just leave as-is for now?

markcoatsworth:
I think this is good to go, over to you to merge.

src/verify_ce_config.py
brianhlin: Yeah, I think it's fine

PR #503, opened 2021-08-02
Fix empty string regex expression warning (HTCONDOR-598)

yongleyuan:


johnkn (commented):
I'm not a python expert.  but I think you just want to build a set of separator characters, consisting of whitespace and ,  and then match one or more of those [, \t]+

yongleyuan:
> I'm not a python expert. but I think you just want to build a set of separator characters, consisting of whitespace and , and then match one or more of those [, \t]+

I have changed the regex and committed it.

brianhlin (approved):
LGTM

PR #505, opened 2021-08-06
Fix condor_ce_run --debug (HTCONDOR-631)

brianhlin:


src/condor_ce_run
```diff
@@ -88,6 +88,19 @@ def parse_opts():
...
 
+
+def run_command_debug(cmd, debug=False):
+    """Run cmd and print stdout/stderr if debug
+    """
+    rc, stdout, stderr = ce.run_command(cmd)
+
+    if debug:
+        print(stdout)
+        print(stderr, file=sys.stderr)
```
markcoatsworth, line 99: Think about printing this to stdout instead of stderr? In some circumstances (ie. piped output, Condor jobs with no error file defined) the stderr output can get dropped.

src/condor_ce_run
```diff
@@ -117,7 +130,7 @@ def submit_job(job_info):
     if job_info['remote']:
         args += ['-remote', job_info['schedd_name'], '-pool', job_info['collector_name']]
 
-    rc, stdout, stderr = ce.run_command(args)
+    rc, stdout, stderr = run_command_debug(args, debug=job_info['debug'])
```
markcoatsworth, line 133: Should check to make sure `job_info['debug']` exists, otherwise this will throw a KeyError. I know it's only called from one place right now where the `debug` key is set, but this is a general-use function that might get used later in other places.

markcoatsworth (approved):
This is a simple change that generally looks good. I've added a couple suggestions you may or may not care about.

src/condor_ce_run
brianhlin: This is just taking the stderr from condor commands like `condor_submit` so it seems appropriate to just pass it through to stderr.

brianhlin:
@markcoatsworth I added a commit to make sure that debug was always defined and improved propagation from the `--debug` option down to the `run_command_debug` function.

markcoatsworth:
All updates look good to me.

markcoatsworth (approved):
All updates look good to me.

PR #506, opened 2021-08-12
Improve SciTokens support in HTCondor-CE submission tools (HTCONDOR-636)

brianhlin:
This PR includes a couple of things:

1.  Adding tests against different OSG repos, which have different versions of HTCondor. The OSG nightly tests have been failing for 3.6 and 3.6 upcoming because we build without GSI, so it was catching some issues that our CI tests weren't.
    - Failing tests: https://github.com/brianhlin/htcondor-ce/actions/runs/1114504850
    - Happy tests with these code changes plus https://github.com/opensciencegrid/osg-test/pull/229: https://github.com/brianhlin/htcondor-ce/actions/runs/1124380171
1.  Changes to submit file generation for `condor_ce_run` to take advantage of the SciTokens commands support as well as initial support for token discovery in `condor_ce_trace` since we construct the job ad ourselves
1.  Simplification of the auth method list to use; users can muck with the list by setting `_condor_SEC_CLIENT_AUTHENTICATION_METHODS` in their environment
1. Minor packaging cleanup for @timtheisen to review


tests/build_rpms.sh
```diff
@@ -70,12 +72,8 @@ fi
...
 mkdir -p /tmp/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
 
-if [[ $BUILD_ENV == uw_build ]]; then
-    printf "%s\n" "%dist .el${OS_VERSION}" >> /etc/rpm/macros.dist
-else
-    printf "%s\n" "%dist .${BUILD_ENV}.el${OS_VERSION}" >> /etc/rpm/macros.dist
-fi
-printf "%s\n" "%${BUILD_ENV} 1" >> /etc/rpm/macros.dist
+printf "%s\n" "%dist .el${OS_VERSION}" >> /etc/rpm/macros.dist
+[[ ${BUILD_ENV} =~ ^osg ]] && printf "%s\n" "%osg 1" >> /etc/rpm/macros.dist
```
markcoatsworth, line 76: This syntax doesn't make sense to me, I didn't think the `[[ ... ]]` was legal without an if statement? What are you trying to do here? Should include a comment.

src/condor_ce_trace
```diff
@@ -316,7 +329,7 @@ def main():
         raise ce.CondorRunException('ERROR: Could not find CE schedd at %s.\n' % job_info['schedd_name'] + \
                                     'Verify that the Scheduler daemon is up with `condor_ce_status -any`.')
 
-    configure_auth()
+    os.environ.setdefault('_condor_SEC_CLIENT_AUTHENTICATION_METHODS', 'SCITOKENS,GSI,FS')
```
markcoatsworth, line 332: If `_condor_SEC_CLIENT_AUTHENTICATION_METHODS` is already set then `os.environ.setdefault()` will not update it with the new value. Do you care?

src/condor_ce_trace
```diff
@@ -195,13 +210,11 @@ def check_job_submit(job_info, schedd_ad):
...
-        if 'SCITOKENS_FILE' not in htcondor.param:
-            raise ce.CondorUserException("Could not find an X509 proxy in %s" % proxy)
-    else:
-        job_ad['x509userproxy'] = proxy
-        add_job_proxy_info(job_ad)
+
+    job_ad.update(setup_user_creds())
+
+    if G_DEBUG:
+        print("Job ad, pre-submit:", job_ad)
```
markcoatsworth, line 217: Use an f-string here for consistency

markcoatsworth (commented):
This generally looks good. I've added a couple of suggestions and confusions, nothing serious.

tests/build_rpms.sh
matyasselmeci: Yep, `[[ ... ]] && foo` is equivalent to `if [[ ... ]]; then foo; fi`

(note that due to precedence, `[[ ... ]] && foo || bar` is _not_ the same as `if [[ ... ]]; then foo; else bar; fi`, it is `if [[ ... ]]; then foo || bar; fi`)

As a side note, a regex is not necessary here... to check if `$BUILD_ENV` starts with `osg`, you can do `[[ $BUILD_ENV == osg* ]]`.

src/htcondorce/tools.py
```diff
@@ -82,3 +83,43 @@ def to_bytes(strlike, encoding="latin-1", errors="backslashreplace"):
...
+def x509_user_proxy_path():
+    """Return the path to the user's X.509 proxy or raise FileNotFoundError if it doesn't exist on disk
+    """
+    try:
+        path = os.environ['X509_USER_PROXY']
+    except KeyError:
+        path = f'/tmp/x509up_u{os.geteuid()}'
+
+    if os.path.exists(path):
+        return path
```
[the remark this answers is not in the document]
matyasselmeci: try doing
```
    if open(path):
        return path
```
which will do all the checks you want and raise the appropriate error if they fail.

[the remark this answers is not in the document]
matyasselmeci: i.e. you will also be able to delete the `raise` statement below

src/condor_ce_trace
brianhlin: Yeah, that's what we want, so users can override the method list if they'd like

tests/build_rpms.sh
brianhlin: > As a side note, a regex is not necessary here... to check if $BUILD_ENV starts with osg, you can do [[ $BUILD_ENV == osg* ]].

Done

src/condor_ce_trace
brianhlin: Fixed

src/htcondorce/tools.py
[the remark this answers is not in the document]
brianhlin: Done

brianhlin:
Force pushed to remove an unused import that I had added + get rid of some trailing whitespace

src/condor_ce_run
```diff
@@ -102,6 +108,12 @@ def generate_submit_contents(job_info):
     else:
         job_info['arguments'] = ''
 
+    try:
+        if ce.x509_user_proxy_path():
+            job_info['use_x509userproxy'] = True
+    except OSError:
+        job_info['use_x509userproxy'] = False
+
```
matyasselmeci, line 116: What's `job_info['use_x509userproxy']` if `ce.x509_user_proxy_path()` is false but doesn't raise an error? Do we have to worry about a KeyError down the line?

matyasselmeci (approved):
Other than the potential KeyError I mentioned, it looks good to me.

src/condor_ce_run
brianhlin: I don't think we have to worry but safety first!

markcoatsworth (approved):
All changes look good to me.

PR #509, opened 2021-09-03
HTCONDOR-662 Fix post-route for CPUs and Memory

johnkn:
Fix bugs and make more consistent with old syntax default route, in particular
     RequestCpus from incoming job is ignored and overwritten unless > 1
     RequestMemory from incoming job is ignored and overwritten
   We do this because CE Admins are reporting that incoming values for RequestCpus
   and RequestMemory are often the minimal defaults set by condor_submit rather than

### DIFF
--- a/config/01-ce-router-defaults.conf
+++ b/config/01-ce-router-defaults.conf
@@ -137,7 +137,6 @@ JOB_ROUTER_TRANSFORM_Cleanup @=jrt
 
 JOB_ROUTER_TRANSFORM_OrigRequests @=jrt
     # Copy original, incoming job attributes for use in post transforms
-    COPY /^Request.+$/             orig_\0
     COPY /^OnExitHold.*$/          orig_\0
     COPY BatchRuntime              orig_BatchRuntime
     COPY environment               orig_environment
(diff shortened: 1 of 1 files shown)


config/01-ce-router-defaults.conf
```diff
@@ -151,89 +150,83 @@ JOB_ROUTER_TRANSFORM_OrigRequests @=jrt
...
+        if $(test_RequestCpus)
+            # RequestCpus already correct
+        else
+            EVALSET RequestCpus $(default_xcount:1)
+        endif
+    endif
+
+    # the old-style transforms ignored RequestCPUs for... reasons? when setting these. we do the same here.
+    EVALSET remote_SMPGranularity  xcount ?: ($(default_xcount:1))
+    COPY    remote_SMPGranularity  remote_NodeNumber
```
brianhlin, line 170: This is a good catch and we can think about fixing this later. I don't believe it's caused any issues to date, though.

config/01-ce-router-defaults.conf
```diff
@@ -151,89 +150,83 @@ JOB_ROUTER_TRANSFORM_OrigRequests @=jrt
...
+
+    SET JobIsRunning           (JobStatus =!= 1) && (JobStatus =!= 5)
+
     if $(test_want_whole_node)
-        SET JOB_GLIDEIN_Cpus       "$$(MY.TotalCpus ?: JobCpus)"
+        SET JOB_GLIDEIN_Cpus       "$$([MY.TotalCpus ?: JobCpus])"
```
brianhlin, line 175: What's the purpose of adding the square brackets here?

config/01-ce-router-defaults.conf
```diff
@@ -151,89 +150,83 @@ JOB_ROUTER_TRANSFORM_OrigRequests @=jrt
...
-    DEFAULT JobCpus                OriginalCpus
-    DEFAULT RequestCpus            OriginalCpus
 @jrt
 
 
 JOB_ROUTER_TRANSFORM_Gpus @=jrt
     # Request GPUs for whole node jobs (HTCONDOR-103)
     # If a whole node job requests GPUs and is matched to a machine with GPUs then set the job's RequestGPUs to all the
     # GPUs on that machine
+    REQUIREMENTS RequestGpus isnt undefined
```
brianhlin, line 194: I think we need something like the following:

```suggestion
    REQUIREMENTS RequestGpus isnt undefined || WantWholeNodes isnt undefined
```

Otherwise `WantWholeNode` requests won't make use of any of the host's GPUs

config/01-ce-router-defaults.conf
```diff
@@ -151,89 +150,83 @@ JOB_ROUTER_TRANSFORM_OrigRequests @=jrt
...
-    # 2. RequestMemory from the source job
-    # 3. default_MaxMemory from the job route
-    # 4. 2000
+    # Outside of the HTCondor WholeNodeJobs case, set RequestMemory to one of the following, in order:
+    # 1. 'maxMemory' from the source job if it's defined
+    # 2. default_MaxMemory from the job route
+    # 3. 2000
+    # Note that if the route does the condor thing and sets RequestMemory, this transform will stomp on that
+
+    EVALSET RequestMemory maxMemory ?: ($(default_maxMemory:2000))
```
brianhlin, line 214: For my own edification, why do we need the outer parens around `$default_maxMemory:2000`?

brianhlin (commented):
LGTM outside of the `REQUIREMENTS` added to the GPU transform

config/01-ce-router-defaults.conf
johnkn: RTM.  $$() syntax is $$(attribute) or $$([expression])

config/01-ce-router-defaults.conf
johnkn: change is needed to make the code do what the comment says it does, your alteration breaks that

config/01-ce-router-defaults.conf
johnkn: default_maxMemory can be an expression here

config/01-ce-router-defaults.conf
brianhlin: For anyone reading this in the future, the documentation @johnkn is referring to can be found in [condor_submit](https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html#index-363), which is perhaps not the first place someone would look for if they're looking for information about job router configuration or ClassAd transforms.

config/01-ce-router-defaults.conf
johnkn: if we give GPUs to jobs that don't want them, it won't be possible to tell from condor_status that the GPUs are going unused

brianhlin: Ah I had thought the comment was wrong but I misread the original logic in `JOB_ROUTER_DEFAULTS` so nevermind on this one

brianhlin (approved):
LGTM

config/01-ce-router-defaults.conf
brianhlin: > if we give GPUs to jobs that don't want them, it won't be possible to tell from condor_status that the GPUs are going unused

OTOH, giving an entire node with a GPU to a pilot that doesn't make use of the GPU is a waste. Though I would guess that such a case is protected against by most `START` expressions?

PR #513, opened 2021-09-10
Ensure default MaxJobs/MaxIdleJobs are the same for all JR syntaxes

brianhlin:
(HTCONDOR-718)

This change makes the default MaxJobs and MaxIdleJobs the same across transform and JRD/JRE syntaxes


johnkn (approved):
looks good

PR #514, opened 2021-09-10
Run osg-test against both UW and osg build CI (HTCONDOR-466)

RyanEweSeng:


tests/test_inside_docker.sh
```diff
@@ -96,10 +54,8 @@ cp /etc/condor/config.d/99-local.conf /etc/condor-ce/config.d/99-local.conf
 # Reduce the trace timeouts
 export _condor_CONDOR_CE_TRACE_ATTEMPTS=60
 
-if [[ $BUILD_ENV == osg* ]]; then
+if [ $BUILD_ENV == osg* ] || [ $BUILD_ENV == uw_build ]; then
```
brianhlin, line 57: @RyanEweSeng we only care about the `osg` and `uw_build` build environments so we can get rid of the if block altogether

brianhlin (commented):
One minor comment but otherwise LGTM as a starting point for someone else to take over. It looks like `osg-test` has some dependencies on `osg-release` that we'll need to tease out: https://github.com/htcondor/htcondor-ce/pull/514/checks?check_run_id=3590126552#step:8:268

tests/test_inside_docker.sh
matyasselmeci: Also, for future reference, if you're writing a bash script (not /bin/sh), then you should avoid using single-bracket expressions (`[ ... ]`) and always use double-bracket because it's safer and has extra features like being able to put the or statement inside the expression. So instead of things like
```
if [ $BUILD_ENV == osg* ] || [ $BUILD_ENV == uw_build ]
```
do
```
if [[ $BUILD_ENV == osg* || $BUILD_ENV == uw_build ]]
```

Again, this is just for future reference; Brian's right that in this case we can get rid of the if block entirely.

timtheisen:
Our CI currently tests against both UW and OSG builds.

brianhlin:
@timtheisen yes but this is for running `osg-test` against both UW + OSG builds. Right now there's some very basic tests that run against the UW builds https://github.com/htcondor/htcondor-ce/blob/V6-branch/tests/test_inside_docker.sh#L19-L59

matyasselmeci:
I think this is going to be moot with HTCSS/OSG 23. Can we close it?

brianhlin:
Agreed, closing

PR #517, opened 2021-10-05
Document common SciTokens problems. HTCONDOR-607

JaimeFrey:


docs/v5/troubleshooting/common-issues.md
```diff
@@ -156,15 +156,83 @@ You may see error messages like the following in your [SchedLog](logs.md#schedlo
...
+
+#### Jobs fail to submit: Verify SSL configuration
+
+Your machine must have a valid host certificate and the CE must be
+configured to use it.
+Both the CE and the submitting machines must be configured to recognize
+the Certificate Authorities (CAs) that issued all certificates in use.
+See the documentation about
+[Configuring Certificates](../configuration/authentication.md#configuring-certificates)
+for details.
```
brianhlin, line 169: I can imagine two different failures here: the server doesn't like the token issuer CA or the client doesn't like the CE's host cert CA. Are there specific errors that we can add here for these cases? What about bad permissions/ownership on the cert and/or key?

docs/v5/troubleshooting/common-issues.md
```diff
@@ -156,15 +156,83 @@ You may see error messages like the following in your [SchedLog](logs.md#schedlo
...
-    LCMAPS_DEBUG_LEVEL=5` to `/etc/sysconfig/condor-ce`, restarting the condor-ce service, and checking
-    `/var/log/messages` for errors again.
+1.  Check the files in `/etc/condor-ce/mapfiles.d/` and ensure that the
+    user's authentication method and identity are present (possibly via a
+    regular expression), and that the mapped OS account exists on your CE
+    and cluster.
+1.  If using GSI, check voms-mapfile or grid-mapfile as an alternate
+    file with a mapping for the user's identity.
+1.  If using lcmaps, check for lcmaps errors in `/var/log/messages`.
+1.  If you do not see helpful lcmaps error messages in `/var/log/messages`,
```
brianhlin, line 232:
```suggestion
1.  If using LCMAPS, check for LCMAPS errors in `/var/log/messages`.
1.  If you do not see helpful LCMAPS error messages in `/var/log/messages`,
```

This was in the old doc but we may as well fix the capitalization while we're in here.

brianhlin (commented):
Some specific comments but it looks pretty solid. What do you think @matyasselmeci ?

PR #519, opened 2021-10-14
Format APEL records with maximum integer length (HTCONDOR-826)

### LINKED ISSUE #518: APEL accounting records may overflow for large memory usage
The ``ImageSize_RAW`` of a job can become excessively large; the ``condor_q -format`` used to create the APEL records only uses ``%d`` formatting, which may overflow and produce negative output for these values.

The negative values are not useful to APEL, and in fact break the automatic accounting reporting completely. See also https://github.com/apel/apel/issues/262.

Record formatting should use a robust(er) format specifier that can handle large(r) values.

maxfischer2781:
This PR changes the printf format specifier used to create APEL records. Previously it was signed int (32 bit) which would wrap around for excessive ``ImageSize_RAW = 79495059964``.

I have changed *all* fields to signed long long int. This is not strictly needed but has no negative impact and seemed more consistent.

Closes #518. FYI @thdesy

### DIFF
--- a/contrib/apelscripts/condor_batch_blah.sh
+++ b/contrib/apelscripts/condor_batch_blah.sh
@@ -62,14 +62,14 @@ do
     batch_record=`TZ=GMT condor_q -job "$file" \
         -format "%s|" GlobalJobId \
         -format "%s|" Owner \
-        -format "%d|" RemoteWallClockTime \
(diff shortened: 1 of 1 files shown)


markcoatsworth (approved):
Simple update, LGTM

PR #521, opened 2021-10-27
BDII correctly raises exception after schedd xquery error (HTCONDOR-743)

markcoatsworth:


brianhlin (commented):
@markcoatsworth this looks good to me but potentially incomplete since we've lost track of the stacktrace from the original support issue (I've been bitten by this in the past so I try to copy these into the ticket description now).

Could you add a try/except to the Schedd obj creation on line 281 catching RuntimeError, dumping the exception to stderr, and raising?

markcoatsworth:
Done. I didn't add this check originally because the code block is not reachable without a condor_schedd, but I forgot there's a race condition between the earlier schedd check and the function call.

contrib/bdii/htcondor-ce-provider
```diff
@@ -274,13 +274,19 @@ def get_cert_issuer(path):
 
 
 def get_ce_jobs(ce_schedd_ad, time_out):
-    ce_schedd = htcondor.Schedd(ce_schedd_ad)
+    try:
+        ce_schedd = htcondor.Schedd(ce_schedd_ad)
+    except RuntimeError as exc:
+        sys.stderr.write("%s\n" % exc)
```
brianhlin, line 280:
```suggestion
        sys.stderr.write("%s: %s\n" % (exc, ce_fqdn))
```

Not sure if the exception contains the CE FQDN so we may as well include it like below

brianhlin (approved):
One suggestion but otherwise LGTM

contrib/bdii/htcondor-ce-provider
markcoatsworth: This isn't going to work as suggested because `ce_fqdn` isn't initialized yet. We can easily bump it up to the top line. This also raises the question: should the `htcondor.Schedd()` and `ce_schedd.xquery()` calls go in the same try/catch since they both catch exceptions exactly the same way?

brianhlin: Good catch. That all sounds good to me!

markcoatsworth: Great! It's all done, going to merge now.

PR #522, opened 2021-10-29
Get the right env var for the token path

matyasselmeci:
BEARER_TOKEN is the contents of the token; BEARER_TOKEN_FILE is the path to the token file


brianhlin (approved):
LGTM

PR #523, opened 2021-11-16
HTCONDOR-598: Document this in the 5.1.2 release

timtheisen:


brianhlin:
I discussed this with TimT in person and this change was added in #503 but the entire function was subsequently ripped out in #506 so we don't have to document this in the changelog

PR #530, opened 2022-02-09
Fix reference to default_CERequirements job attr (HTCONDOR-964)

brianhlin:
Without this, users always get:

CERequirements = "default_CERequirements,CondorCE"


brianhlin:
EL8 tests failed, potentially due to a mirror issue: https://github.com/htcondor/htcondor-ce/runs/5133291016?check_suite_focus=true#step:5:34

Re-kicking tests in hopes that it was indeed transient

matyasselmeci:
The EL8 images are still using CentOS Linux 8 (not Stream) as a base and the repos for that are gone.

brianhlin:
The GHA interface didn't have a way for me to re-run the test (presumably because the workflow was old?) so I added a dummy commit and removed it to kick off a new round, hopefully with the fixed tests in #533

PR #531, opened 2022-02-16
default_CERequirements needs MY. in front of it

lebonez:


lebonez:
whoops didn't realize this was already put in my Brian

PR #535, opened 2022-03-26
Use integer conversion for Cpus and Gpus

djw8605:
The CPUs and GPUs are floats, for some reason.  Convert to integer
which makes it easier to parse with python classad bindings (for example
in the gratia probe)

The gratia probe reads `MATCH_EXP_JOB_GLIDEIN_Cpus`.

I am open to opinions.


timtheisen:
So, in the near future we will be allocating fractional GPUs. And, allocating fractional CPUs in on the road map. What is the difficulty in parsing with the Python ClassAd bindings? Is there a bug there that we need to fix?

PR #536, opened 2022-04-05
HTCONDOR-897 Docs for client auth in AuditLog

JaimeFrey:
Add docs for checking authentication method and identity of job submitters.


brianhlin (approved):
LGTM!

PR #539, opened 2022-05-11
HTCONDOR-1085 Remove setting default values of GSI config params

JaimeFrey:
Don't explicitly set the GSI-related parameters to their default values.
This doesn't change HTCondor's behavior, but does alter the environment,
which can leak to tools like slurm's sbatch. sbatch will forward those
environment variables to the user job.


brianhlin (approved):
LGTM!

PR #540, opened 2022-05-13
HTCONDOR-1017 Rename source job AuthToken* attributes in CE routed jobs

johnkn:


brianhlin (changes requested):
We also need `JOB_ROUTER_DEFAULTS` updated as we haven't removed support for that yet and I believe a fair amount of CEs are still using the old config style

brianhlin (approved):
LGTM!

PR #543, opened 2022-06-07
HTCONDOR-1153: Incorporate downstream OSG python3-rrdtool change

timtheisen:


edquist (approved):
From the osg perspective i think the changes look fine.  Yes in osg we build a python3-rrdtool.  Only, i thought upstream did also, so i am not sure whether or not rrdtool by itself will give you the python pieces you want.

timtheisen:
This is a null change for upstream. Are you saying that we need to build python3-rrdtool for HTCondor-CE UW-build to be useful?

Also, OSG builds and tests failed on EL7. Perhaps, I need to adjust the tests.

timtheisen:
Does the HTCondor-CE work OK with the regular rrdtool for the UW-build. Carl raised the question.

brianhlin:
It should be fine. rrdtool is used for `htcondor-ce-view` and falls back to the CLI tool if the python bindings aren't available. I don't imagine that we'll have the same efficiency issues that the GlideinWMS factory was seeing with this fallback as we don't have to deal with the same scale of info

PR #544, opened 2022-06-09
(HTCONDOR-1206) Apel: Default and spec scaling

### LINKED ISSUE #542: Simplify APEL accounting: Average and per-machine performance
Simplify APEL accounting: Average and per-machine performance

Currently APEL resource usage reporting for HTCondor-CE work via an average performance absolute score on the CE and an individual performance scaling factor on each worker node. There are some issues with this approach:
- In a homogeneous cluster, one still needs to configure a "dummy" scaling factor of 1 for each worker node.
- In a heterogeneous cluster, one must publish the average to each worker node to compute its scaling factor.

The former is quite a hassle that makes a minimal setup needlessly complicated. The latter is especially bad in large and/or opportunistic clusters since the average can change frequently; accurately maintaining both average score and per-machine factor at scale is next to impossible. FWIW, it looks like one can in principle disable scaling for the entire cluster but this seems to be undocumented.

I propose to make the homogeneous case simpler to configure and the heterogeneous case simpler to get right but would like feedback before drafting a PR.

-----------

From the APEL side, we have
* an average performance score for the CE itself of a specific type (e.g. `HEPSPEC`) and value (e.g. `13.45`) in `client.cfg`:
   ```
   manual_spec1 = htcondor-ce-1-kit.gridka.de:9619/htcondor-ce-1-kit.gridka.de-condor,HEPSPEC,13.45
   ```
* an [*optional* scaling factor for each job defaulting to 1.0](https://github.com/apel/apel/blob/ef23b80716da51d44a18a82ebf594faaa2e50cf0/apel/parsers/htcondor.py#L47-L49) in each job record.

APEL takes care of matching the average performance to the scale factor (actually, the job) based on the definition time of the `manual_specX` and the end-time of the job.

For the homogeneous case what APEL provides is already sufficient: When creating job records, **if no scaling is available either skip the scaling factor or set it to 1.**
* Should there be an option to force scaling/performance to be set, e.g. to detect configuration errors?

For the heterogeneous case, it would be more robust to **compute the scale factor on the CE**: The StartD/Job would supply an absolute performance and when creating job records **the script would compute the scale factor from the StartD/Job performance and the APEL performance.**
* Should there be a way to provide different performance types (`HEPSPEC`, `SI2K`, ...)?

maxfischer2781:
This PR rewrites the APEL batch/blah script to simplify configuration for the performance scaling factor. Major changes include:

* [x] Port the batch/blah script to Python to simplify changes.
   * [x] Add a `--dry-run` flag to preview what the script does.
* [x] Modify how the performance scaling factor is derived.
   * [x] Scaling may be *computed* from a job's StartD `ApelSpec` attribute and the local APEL client config. This removes the need to publish and synchronise the average performance to all worker nodes.
   * [x] Scaling may be *defined* from a job's StartD `ApelScaling` attribute. This is the current behaviour.
   * [x] Scaling defaults to `1.0` (configurable). This removes the need to adjust worker nodes when performance is homogeneous.
   * Generally, the changes trade ease of use to error detection. An undefined setting is ignored instead of leading to an error.

Closes #542.

---------

The computed scaling relies on a new StartD attribute `ApelSpecs`. This must be a ClassAd which maps from APEL spec types to benchmark values.
```
ApelSpecs = [HEPSPEC=14.37; SI2K=2793]
STARTD_ATTRS = $(STARTD_ATTRS) ApelSpecs
```
The spec types are case *insensitive*. Quoted names are supported, e.g. `'HEP SPEC'`.

The batch/blah script generates an appropriate ClassAd expression from the APEL client config and uses that in its `condor_q` calls for each job. This goes through the `manual_specN` settings in-order and filters for the appropriate `APEL_CE_ID`. If the client defines both `HEPSPEC` and `SI2K`, the generated expression looks like this:
```
((MachineAttrApelSpecs0.'HEPSPEC' / 13.45) ?: ((MachineAttrApelSpecs0.'SI2K' / 2614.0) ?: (MachineAttrApelScaling0 ?: 1.0)))
```


maxfischer2781:
I am currently testing these changes in a production setup, will update the PR once I'm sure things are working.

maxfischer2781:
@brianhlin Thanks for the ping and assigning a review.

Yes, I've just finished testing and everything seems to work with the current status. It would be good to have a second pair of eyes check the various configuration knobs to use consistent names.

I haven't been able to find out why the RPM build fails, though. :/

contrib/apelscripts/condor_batch_blah.py
```diff
@@ -0,0 +1,170 @@
...
+        scale_query = f"({scaling_attr} ?: {scale_query})"
+    try:
+        spec_attr = read_ce_config_val("APEL_SPEC_ATTR")
+    except CalledProcessError:
+        return scale_query
+    else:
+        spec_scale_query = scale_query
+        for spec_type, spec_value in reversed(list(specs.items())):
+            spec_scale_query = (
+                f"(({spec_attr}.'{spec_type}' / {spec_value}) ?: {spec_scale_query})"
```
Todd-L-Miller, line 89: Maybe I'm missing something here, but this seems like it will have the net effect of setting `spec_scale_query` to the first (and only the first) item listed in each CE's spec entry.  If that's the desired behavior, why loop over the whole list?

Todd-L-Miller: The code otherwise looks fine, although I also can't tell why the RPM build fails.

timtheisen: The RPM build fails on one platform, because we are trying to build for an old unsupported platform. If the review is approved, I'll use my admin privileges to merge it. I'll modify the tests to remove the obsolete ones.

maxfischer2781: The expression is intended to go through the different spec types of the CE and use the first one which was defined on the resource as well: the `{spec_attr}.'{spec_type}'` part is undefined if the resource does not define the spec type.

Let's say the expression built is
```
((MachineAttrApelSpecs0.'HEPSPEC' / 13.45) ?: ((MachineAttrApelSpecs0.'SI2K' / 2614.0) ?: (MachineAttrApelScaling0 ?: 1.0)))
```
and the machine only defines `ApelSpecs = [SI2K=2793]`. Then the entire `(MachineAttrApelSpecs0.'HEPSPEC' / 13.45)` part is `undefined` and Elvis picks the part checking `SI2K` instead.

I'll add a comment into the code to explain this.

Todd-L-Miller (approved):

contrib/apelscripts/condor_batch_blah.py
Todd-L-Miller: Sorry, I misread the code.  I didn't understand that it was extending `spec_scale_query` by one clause each time; I thought for some reason it was replacing `spec_scale_query` on each iteration.  The changes look good.

timtheisen (approved):
LGTM. I'll submit a follow-up PR to address the build failures.

PR #546, opened 2022-07-15
HTCONDOR-1211: Fix str/bytes conversion issue in CEView

matyasselmeci:


matyasselmeci:
I live-patched collector-itb.opensciencegrid.org and collector.opensciencegrid.org (in that order, don't worry!) and things seem to work fine.

PR #547, opened 2022-07-27
HTCONDOR-1206: Fix failed builds

timtheisen:
I fixed up the build failures. These are purely mechanical changes.
- osg-3.5-upcoming is no longer supported, dropped from build
- package additional .pyc files for apel script (on EL7)

### DIFF
--- a/.github/workflows/build_and_test_rpms.yml
+++ b/.github/workflows/build_and_test_rpms.yml
@@ -12,7 +12,6 @@ jobs:
           - rockylinux:8
         target_env:
           - uw_build
-          - osg-3.5-upcoming
           - osg-3.6
           - osg-3.6-upcoming
     steps:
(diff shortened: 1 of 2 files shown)


timtheisen:
Actually, this is more of a packaging issue. I'll get it to Mat instead.

matyasselmeci:
So you didn't see the `__pycache__` files on EL 8? That's surprising -- it's something that Python 3 creates but not Python 2.

timtheisen:
If you look at other sub packages you'll see the same pattern. The EL7 builds produce .pyc files and the EL8 builds do not.

matyasselmeci (approved):

matyasselmeci:
Huh. OK, well if that's what you're seeing, that's how it is :-)

PR #549, opened 2022-09-14
HTCONDOR-1326: Use the C++ collector plugin

timtheisen:


Todd-L-Miller (approved):
Looks good to me.

brianhlin:
According to the ticket, we need to wrap these changes in an `if` or force sites to update to at least HTCondor 9.0.1. Also, if it's the latter, can't we remove the old payload audit bits?

timtheisen:
@brianhlin The C++ plugin is a drop in replacement for the Python plugin. So, I don't know of any bits to remove. Rather than complicate the configuration with conditionals on more than one file, I'd rather just require a later condor. I'll bump the condor requirement in the RPM.

PR #551, opened 2022-09-15
HTCONDOR-1326: Need condor 9.0.1 for C++ Collector plugin

timtheisen:


brianhlin (approved):
LGTM, though I'd still like to see us rip out the Python payload auditing it if HTCSS isn't going to support it any longer: https://github.com/htcondor/htcondor-ce/blob/V5-branch/src/htcondorce/audit_payloads.py

brianhlin (approved):
Ya, LGTM!

PR #552, opened 2022-09-30
HTCONDOR-1344 update security config

JaimeFrey:
Update security configuration (use of authentication, encryption, integrity, auth methods) to match recommended settings in HTCondor.


brianhlin (changes requested):
I think the `security:recommended_v9_0` template may come with enough extra stuff that it's probably not what we want. For example, the following `ALLOW_*` knobs are all previously set in https://github.com/htcondor/htcondor-ce/blob/V5-branch/config/01-ce-auth-defaults.conf#L30-L41:

```
ALLOW_ADMINISTRATOR = $(1:condor)@* condor_pool@* $(2:root@*) $(3+)
ALLOW_DAEMON = $(1:condor)@*  condor_pool@*
ALLOW_NEGOTIATOR = $(1:condor)@* condor_pool@*
ALLOW_WRITE = *
ALLOW_READ = *
```

The existing config in `01-ce-auth-defaults.conf` could probably use a review but I think it's out of scope for this ticket.

Additionally, does the default `SEC_CLIENT_AUTHENTICATION_METHODS` list match what we're removing here? I'd be surprised if `SCITOKENS` and `GSI` are in the default list.

Note that there are two different, products that we have to worry about in terms of auth configuration: the CE itself and the central collector (https://htcondor.com/htcondor-ce/v5/installation/central-collector/). They both share `01-common-auth-defaults.conf` but the CE also uses `01-ce-auth-defaults.conf` and the central collector uses `05-ce-collector-auth.conf`. WRT to `SEC_*_{AUTHENTICATION,ENCRYPTION,INTEGRITY}`, I believe it should be safe to set those all to `REQUIRED` for the CE. OTOH, the central collector should probably use the 9.0 recommended defaults of `OPTIONAL` for `READ` because we force clients to present an SSL certificate. I didn't make this clear in the ticket so that's my bad!

My suggested course of action:

1. Move `01-ce-auth-defaults.conf` to `05-ce-auth-defaults.conf` so that `01-common-auth-defaults.conf` are parsed first
2. Remove all of the authentication, encryption, and integrity knobs from the common auth config (I'm assuming that the HTCondor defaults are all `REQUIRED`)
3. Remove `AUTH_SSL_REQUIRE_CLIENT_CERTIFICATE = True` from`05-ce-collector-auth.conf`. This was added for https://opensciencegrid.atlassian.net/browse/HTCONDOR-344, which should no longer be necessary in the post-GSI world and AFAIK OSG has the only deployed central collector.

@timtheisen all of these changes have the potential fo causing pain. Should we move this work to a `V6-branch` (or whatever @tannenba wants to use for CE versioning)?

JaimeFrey:
Good catch on the ordering of 01-ce-auth-defaults.conf and 01-common-auth-defaults.conf. I had it reversed in my head while comparing the settings.
SCITOKENS has been in the default auth methods since 8.9.12. Versions 9.1.3-9.2.0 include GSI support but don't have it in the default auth methods. Can we ignore these releases?

JaimeFrey:
The default for authentication, encryption, and integrity is OPTIONAL. If the server and client agree on AES for a crypto method, then encryption and integrity become REQUIRED. But a client is free to say it only wants BLOWFISH or 3DES.

JaimeFrey:
If the central collector doesn't have the 01-ce-auth-defaults.conf config file, then how is ALLOW_READ set for it? Can only ADMINISTRATOR users query it?

JaimeFrey:
Ah. Both `htcondor-ce` and `htcondor-ce-collector` include `01-ce-auth-defaults.conf`.

JaimeFrey:
I'll make an update to the PR that tries to address all of the issues we've discussed here.

config/05-ce-auth-defaults.conf
```diff
@@ -40,13 +39,13 @@ C_GAHP.ALLOW_DAEMON = $(ALLOW_DAEMON)
...
-# These security settings are likely unnecessary as we don't run a negotiator or
-# a startd.  However, we'd prefer to see the least number of authorizations possible.
+# This security setting is likely unnecessary as we don't run a negotiator.
+# However, we'd prefer to see the least number of authorizations possible.
 ALLOW_NEGOTIATOR = $(SUPERUSERS)
-ALLOW_OWNER = $(SUPERUSERS)
 
 # Authentication settings
 SEC_DEFAULT_AUTHENTICATION_METHODS = FS
+SEC_CLIENT_AUTHENTICATION_METHODS = FS, TOKEN, SCITOKENS, GSI, SSL
```
brianhlin, line 48:
```suggestion
```

config/01-common-auth-defaults.conf
```diff
@@ -13,18 +13,7 @@
...
-SEC_CLIENT_AUTHENTICATION_METHODS = FS, TOKEN, SCITOKENS, GSI, SSL
-
-# Privacy settings
-SEC_DEFAULT_ENCRYPTION = OPTIONAL
-SEC_DEFAULT_INTEGRITY = REQUIRED
-SEC_READ_INTEGRITY = OPTIONAL
-SEC_CLIENT_INTEGRITY = OPTIONAL
-SEC_READ_ENCRYPTION = OPTIONAL
-SEC_CLIENT_ENCRYPTION = OPTIONAL
+use security:recommended_v9_0
```
brianhlin, line 16:
```suggestion
SEC_CLIENT_AUTHENTICATION_METHODS = FS, TOKEN, SCITOKENS, GSI, SSL
use security:recommended_v9_0
```

brianhlin (changes requested):
If we move the `SEC_CLIENT_AUTHENTICATION_METHODS` into the CE auth file, then the `htcondor-ce-client` package won't have the same client auth method list. Otherwise this LGTM.

We may also need the PR re-opened against a new branch, depending on the version number that @timtheisen and @tannenba decide this should go into.

config/01-common-auth-defaults.conf
JaimeFrey: The default auth method list (as in when neither CLIENT nor DEFAULT param is set) is the same as this explicit value (except it includes CLAIMTOBE, which the server can reject).

brianhlin: Gotcha, that makes sense!

brianhlin:
This LGTM but I'm going to leave my review in "Requested changes" depending on the outcome of the version discussion.

JaimeFrey:
Sounds good. We can change the merge-to target of the PR to a new V6-branch without changing the point it was branched from (presuming the new V6-branch is made downstream from where the PR was branched).

timtheisen:
@brianhlin This PR has been retargeted to the V6-branch. OK to merge?

brianhlin (approved):
Yup, LGTM now!

PR #555, opened 2022-10-21
HTCONDOR-1115: Initial draft of the test token generator

bbockelm:
This utility, when run as user `condor` on the CE, will generate a token for an arbitrary issuer / subject combination.

This works by generating a new keypair and inserting the public key into the public key cache; then, the new token is signed by the transient keypair.

The token should be recognized by the CE until the next cache expiration (typically a few hours).

### DIFF
--- a/src/condor_ce_test_token
+++ b/src/condor_ce_test_token
@@ -1,0 +1,215 @@
+#!/usr/bin/python3
+
+import argparse
+import base64
+import binascii
+import json
+
+# For generation of the transient keypair
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives import hashes, serialization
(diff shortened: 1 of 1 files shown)


matyasselmeci (approved):
Pretty nifty! Can you fix the pylint/flake8 errors? You might have to push a dummy commit or something to rerun the checks since the logs have been deleted.

tannenba:
Assigned to Jaime....

tannenba:
Also, wondering if this should be in the htcondor repo instead of the htcondor-ce repo (i.e. is it useful for user communities outside of just the htcondor-ce) ?

JaimeFrey:
We're moving this code to the HTCondor code base (with modifications).

PR #557, opened 2023-01-10
HTCONDOR-1522 Remove some hard-coded directory permission checks

JaimeFrey:
These existance/permission checks assume the admin hasn't altered the default locations, and least SPOOL is likely to ben relocated. The condor_master will create these directories if necessary, in the locations given in the configuration files.
Leaving the checks for LOG and LOCK, since the master needs to write to its log file before creating these directories and the user subdirectories are not created by the master.


timtheisen:
Didn't the ticket request to use "condor_ce_config_val LOG" to get the location of the log directory?

timtheisen (approved):
LGTM

PR #560, opened 2023-01-31
Try double backslashes in mapfile

timtheisen:


timtheisen:
This broke the mapping and had no effect on the test results.

PR #562, opened 2023-02-06
HTCONDOR-1579: Revert to matching ' ' instead of [:space:]

timtheisen:
This avoids a bug with PCRE2 on EL7, where a hyphen following a POSIX character set in a character class is always illegal. This should be legal when the hyphen is the last character in the character set.


matyasselmeci (approved):
Does what it says on the can

matyasselmeci:
What's with the test failures? I've seen them on more than one PR...

brianhlin:
There's some undiagnosed problem with 9.12.0 + HTCondor-CE (maybe just v6?). Not sure how much we care about it given that 9.12.0 isn't technically supported with 10.0 out

PR #572, opened 2023-03-17
HTCONDOR-1651 Overhaul SSL certificate configuration

JaimeFrey:
Now, the base configuration works out-of-the-box with either the standard or grid certificate file locations.
Also, disable and ignore the collector's autogenerated CA and host cert.

### DIFF
--- a/config/01-ce-auth.conf
+++ b/config/01-ce-auth.conf
@@ -7,22 +7,6 @@
 #
 ###############################################################################
 
-# Uncomment the following lines if your SciTokens or SSL clients use
-# grid certificates or your HTCondor-CE's host certificate is located
-# in the standard grid location
-#
-# https://htcondor-ce.readthedocs.io/en/latest/installation/htcondor-ce#configuring-certificates
-
-# AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/hostcert.pem
-# AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/hostkey.pem
-# AUTH_SSL_SERVER_CADIR = /etc/grid-security/certificates
-# AUTH_SSL_SERVER_CAFILE =
-# AUTH_SSL_CLIENT_CERTFILE = /etc/grid-security/hostcert.pem
-# AUTH_SSL_CLIENT_KEYFILE = /etc/grid-security/hostkey.pem
-# AUTH_SSL_CLIENT_CADIR = /etc/grid-security/certificates
-# AUTH_SSL_CLIENT_CAFILE =
-
-
 # By default, regular expressions in the second field of HTCondor-CE
 # mapfiles must be enclosed with '/'. For exmaple:
 #
--- a/config/01-common-auth-defaults.conf
+++ b/config/01-common-auth-defaults.conf
@@ -21,8 +21,16 @@ SEC_PASSWORD_DIRECTORY = /etc/condor-ce/passwords.d
 # GSI settings
 CERTIFICATE_MAPFILE=/etc/condor-ce/condor_mapfile
 
-# SSL settings, appropriate for non-grid hosts
-AUTH_SSL_SERVER_CAFILE = /etc/pki/tls/certs/ca-bundle.crt
+# Alter SSL settings to work with both standard and grid file locations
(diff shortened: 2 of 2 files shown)


config/01-ce-auth.conf
```diff
@@ -7,22 +7,6 @@
...
-# Uncomment the following lines if your SciTokens or SSL clients use
-# grid certificates or your HTCondor-CE's host certificate is located
-# in the standard grid location
-#
-# https://htcondor-ce.readthedocs.io/en/latest/installation/htcondor-ce#configuring-certificates
-
-# AUTH_SSL_SERVER_CERTFILE = /etc/grid-security/hostcert.pem
-# AUTH_SSL_SERVER_KEYFILE = /etc/grid-security/hostkey.pem
-# AUTH_SSL_SERVER_CADIR = /etc/grid-security/certificates
-# AUTH_SSL_SERVER_CAFILE =
```
brianhlin, line 19: I would actually prefer keeping this. We still want to use the OS CAs by default when we only have the option of picking one or the other instead of ranking them

config/01-ce-auth.conf
```diff
@@ -7,22 +7,6 @@
...
-# AUTH_SSL_CLIENT_CERTFILE = /etc/grid-security/hostcert.pem
-# AUTH_SSL_CLIENT_KEYFILE = /etc/grid-security/hostkey.pem
-# AUTH_SSL_CLIENT_CADIR = /etc/grid-security/certificates
-# AUTH_SSL_CLIENT_CAFILE =
```
brianhlin, line 23: Same comment here

config/01-common-auth-defaults.conf
```diff
@@ -21,8 +21,16 @@ SEC_PASSWORD_DIRECTORY = /etc/condor-ce/passwords.d
 # GSI settings
 CERTIFICATE_MAPFILE=/etc/condor-ce/condor_mapfile
 
-# SSL settings, appropriate for non-grid hosts
-AUTH_SSL_SERVER_CAFILE = /etc/pki/tls/certs/ca-bundle.crt
+# Alter SSL settings to work with both standard and grid file locations
+AUTH_SSL_SERVER_CERTFILE = /etc/pki/tls/certs/localhost.crt,/etc/grid-security/hostcert.pem
+AUTH_SSL_SERVER_KEYFILE = /etc/pki/tls/private/localhost.key,/etc/grid-security/hostkey.pem
+AUTH_SSL_SERVER_CADIR = /etc/grid-security/certificates
```
brianhlin, line 27:
```suggestion
```

config/01-common-auth-defaults.conf
```diff
@@ -21,8 +21,16 @@ SEC_PASSWORD_DIRECTORY = /etc/condor-ce/passwords.d
...
+AUTH_SSL_CLIENT_CERTFILE = /etc/pki/tls/certs/localhost.crt,/etc/grid-security/hostcert.pem
+AUTH_SSL_CLIENT_KEYFILE = /etc/pki/tls/private/localhost.key,/etc/grid-security/hostkey.pem
+AUTH_SSL_CLIENT_CADIR = /etc/grid-security/certificates
```
brianhlin, line 30:
```suggestion
```

brianhlin (changes requested):
Minor changes requested.

config/01-ce-auth.conf
JaimeFrey: With this configuration, the code should consider both the OS CAs and the grid CAs. You can provide both a CA file and a CA directory to OpenSSL, and it should look in both for any request.

brianhlin: Oh, that's good! What wins if there's a collision, i.e. Let's Encrypt in both the CA file + CA dir?

config/01-common-auth-defaults.conf
brianhlin: Nevermind, we can do both. Cool!

config/01-ce-auth.conf
brianhlin: The question was more of a curiosity -- we don't actually know what happens -- but I don't think we should hold up this change for my question!

brianhlin (approved):
LGTM!

PR #574, opened 2023-04-04
HTCONDOR-1663 spool-config-check

JaimeFrey:
When routing into a local HTCondor pool, verify that the job router will find the local schedd's SPOOL directory.


brianhlin (approved):
LGTM! However, would you be able to update the comment on L157 to note the additional checks in this block? It's far enough outside of the diff that GitHub doesn't give me the option to make a suggestion :( 

https://github.com/htcondor/htcondor-ce/pull/574/files#diff-0d251899238b2c7e9506311f6fc0332b7afaf20bba87aca36510c6b2407ebf96R157

PR #575, opened 2023-04-12
HTCONDOR-1657 restrict NEGOTIATOR auth level to FS

johnkn:


config/05-ce-auth-defaults.conf
```diff
@@ -33,7 +33,7 @@ SCHEDD.ALLOW_WRITE = $(USERS), condor@daemon.htcondor.org/$(FULL_HOSTNAME)
 COLLECTOR.ALLOW_ADVERTISE_MASTER = $(FRIENDLY_DAEMONS)
 COLLECTOR.ALLOW_ADVERTISE_SCHEDD = $(FRIENDLY_DAEMONS)
 COLLECTOR.ALLOW_ADVERTISE_STARTD =  $(UNMAPPED_USERS), $(USERS)
-SCHEDD.ALLOW_NEGOTIATOR = condor@daemon.htcondor.org/$(FULL_HOSTNAME)
+SCHEDD.ALLOW_NEGOTIATOR = condor@$(UID_DOMAIN)
```
brianhlin, line 36:
```suggestion
SCHEDD.ALLOW_NEGOTIATOR = condor@$daemon.htcondor.org
```

The `@daemon.htcondor.org/$(FULL_HOSTNAME)` may not be ideal as we've discussed (maybe there's a ticket for it?) but I don't think `UID_DOMAIN` would be right either due to:

- https://github.com/htcondor/htcondor-ce/blob/V6-branch/config/05-ce-auth-defaults.conf#L22-L22
- https://github.com/htcondor/htcondor-ce/blob/V6-branch/config/mapfiles.d/50-common-default.conf#L13

brianhlin (commented):
I think the `ALLOW` value has to be fixed, otherwise this change does what it says on the tin. I'm just not entirely sure why it's needed?

PR #577, opened 2023-04-14
Add the ability to attempt submission without a credential (HTCONDOR-1787)

brianhlin:
Useful for admins to test CE -> local batch submission without having to worry about the SciToken credentials

### DIFF
--- a/src/condor_ce_trace
+++ b/src/condor_ce_trace
@@ -87,6 +87,8 @@ def parse_opts():
                       dest="schedd_name")
     parser.add_option("-d", "--debug", help="Print debugging info.",
                       dest="debug", default=False, action="store_true")
+    parser.add_option("--skip-scitokens", help="Skip SCITOKENS authentication.",
+                      dest="skip_scitokens", default=False, action="store_true")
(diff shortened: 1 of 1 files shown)


brianhlin:
Tried this out on the canary Hosted CE and it worked

johnkn (approved):
commit message should reference the JIRA ticket.  other than that it looks ok to me.

brianhlin:
Fixed commit message but RPM builds on EL8 seem to be failing for an unrelated reason.

@timtheisen how would you like to proceed?

matyasselmeci:
The error is in the `git archive` call. Newer versions of git refuse to do certain operations on repos not owned by the current user:
```
+ git archive --format=tar --prefix=htcondor-ce-6.0.0/ HEAD
+ gzip
fatal: detected dubious ownership in repository at '/htcondor-ce'
```
since `/htcondor-ce` is bind-mounted from outside of the container, that is probably what is happening.  Fortunately, git tells you what to do:

```
To add an exception for this directory, call:

	git config --global --add safe.directory /htcondor-ce
```

PR #583, opened 2023-09-12
HTCONDOR-2030: Fix tests

timtheisen:


timtheisen:
Just having the test run properly is probably all that is needed. However, the reviewer should ensure that I did not get things to pass via nefarious means (such as removing tests or ignoring failures).

ColeBollig (approved):
LGTM... no nefarious bypassing

PR #587, opened 2023-09-21
Fix RPM upload GHA

brianhlin:


brianhlin:
FWIW, we probably should just not use `/usr/sbin/init` as PID 1 in the container. I don't think it really matters what proc it is so we could probably just use `/bin/sh` but that's something that can be left for later

matyasselmeci:
Why do you need --privileged?

matyasselmeci:
You're not starting any services. Don't even specify a command to _docker run_; you'll get the default shell.

timtheisen:
Otherwise the EL9 container does not stay running.

matyasselmeci:
> Otherwise the EL9 container does not stay running.

What is the error message you get?

timtheisen:
`Error response from daemon: Container 497247badd539902a91fab1a202f33dcb8b108303e8f8c1348c1ac0ede441b6d is not running`

PR #588, opened 2023-09-29
Preliminary HTCondor-CE 23 docs

timtheisen:


brianhlin (approved):
LGTM other than a dead link in `v6/releases.md` that points to the now removed `v5/releases.md`

PR #591, opened 2023-11-16
HTCONDOR-2151 unmapped-users

JaimeFrey:


config/mapfiles.d/50-central-collector.conf
```diff
@@ -1,11 +0,0 @@
...
-#
-# HTCondor-CE Central Collector default authentication mappings
-#
-# DO NOT EDIT THIS FILE!  It will be overwritten upon RPM upgrade.
-# If you wish to make changes to the HTCondor-CE mappings, create files
-# in /etc/condor-ce/mapfiles.d containing your changes.
-#
-###############################################################################
-
-GSI /[.A-Za-z0-9\/= -]*\/CN=([.A-Za-z0-9\/= -]+)/ \1@unmapped.htcondor.org
```
brianhlin, line 11: If GSI does get used, do unmapped users get mapped to `gsi@unmapped`?

brianhlin (approved):
One question but otherwise looks reasonable to me

config/mapfiles.d/50-central-collector.conf
JaimeFrey: If HTCondor 9.0.0 is being used, then yes. I hope we can assume a more recent version of HTCondor at this point.

brianhlin: Yeah, you guys don't support anything older so I think this is kosher!

brianhlin:
@timtheisen does this change need to be backported to HTCondor-CE 6? I forget the support policy details

timtheisen:
This would need to be back ported if it is security issue (CVE) or a critical bug fix. Otherwise, no back-port is needed.

JaimeFrey:
I have no concrete security or bug concern here.

PR #592, opened 2023-12-19
HTCONDOR-2193 gpu-reqs

JaimeFrey:
Condor-C jobs won't have this check added automatically and startds that don't have GPU detection won't include a check. If we don't add a check here, then jobs with RequestGPUs set can end up running on machines with no GPUs when routed to a local HTCondor pool.

We only set Requirements if it's currently the literal value 'true'. We expect newer clients to start setting all of the automatic requirements clauses for Condor-C jobs, and we don't want to ignore those.


johnkn (approved):
LGTM

PR #598, opened 2024-02-13
Disable GSI in 05-ce-auth-defaults.conf

matyasselmeci:
GSI no longer works in HTCondor so these just cause an annoying warning.


JaimeFrey:
There's more GSI cruft to be ripped out, but this addresses things that hit users in the face.

JaimeFrey (approved):
LGTM

PR #599, opened 2024-03-05
HTCONDOR-2322 job-env-transform

JaimeFrey:
The macros CONDORCE_PILOT_JOB_ENV and default_pilot_job_env should already include quotes, so adding quotes results in a ClassAd syntax error. default_pilot_job_env may be undefined, so set a default of "".


ColeBollig (approved):
LGTM

PR #600, opened 2024-04-05
superusers are also queue superusers

maxfischer2781:
This PR fixes the permissions of the `root` and `condor` users for queue commands.

Currently, `QUEUE_SUPER_USERS = condor, root` lacks a domain so the `UID_DOMAIN = users.htcondor.org` is implied. However, the [default mapping for local `root`/`condor` users](https://github.com/htcondor/htcondor-ce/blob/729edd27f5c1f25f2926a04cb5c6fd2fd1433f45/config/mapfiles.d/50-common-default.conf#L12) uses the domain `daemon.htcondor.org` instead. As such, `root` is not allowed to perform administrative queue commands such as `condor_ce_rm`.

### DIFF
--- a/config/05-ce-auth-defaults.conf
+++ b/config/05-ce-auth-defaults.conf
@@ -37,7 +37,7 @@ SCHEDD.ALLOW_NEGOTIATOR = condor@daemon.htcondor.org/$(FULL_HOSTNAME)
 ALLOW_DAEMON = $(FRIENDLY_DAEMONS)
 C_GAHP.ALLOW_DAEMON = $(ALLOW_DAEMON)
 ALLOW_ADMINISTRATOR = $(SUPERUSERS)
-QUEUE_SUPER_USERS = condor, root
+QUEUE_SUPER_USERS = $(SUPERUSERS)
 
 # This security setting is likely unnecessary as we don't run a negotiator.
 # However, we'd prefer to see the least number of authorizations possible.


maxfischer2781:
I've just traced back why the change occurred, since it did not appear on a v5.1.6 CE but on a new v23 CE even though both tags include the mapfile. [The mapfile is only installed with the client tools](https://github.com/htcondor/htcondor-ce/blob/729edd27f5c1f25f2926a04cb5c6fd2fd1433f45/rpm/htcondor-ce.spec#L487) which we do have on the new CEs (for local automated testing) but not the older ones.  
So `root`/`condor` get mapped differently depending on what other packages are installed. Consequently, different `QUEUE_SUPER_USERS` are needed depending on the situation.

I think it's not too unusual to have the client installed on a CE, since it is recommended for troubleshooting. So ideally the mapping should be the same both with and without the client installed.

Would you be fine with moving the mapfile to the regular installation as well? Otherwise I would recommend removing it and keeping the old `QUEUE_SUPER_USERS` to avoid the conflict from installing the client.

maxfischer2781:
Sorry for the rambling. The client is always installed with the CE and always has been. I've just double-checked and the issue also occurs on our older machines, we just always interacted with the LRMS instead of CE queue.

JaimeFrey:
I'm surprised this change works. The QUEUE_SUPER_USERS parameter doesn't support a `/$(FULL_HOSTNAME)` at the end of entries.
It seems like we could just get rid of the 50-common-default.conf mapfile from the client package.

maxfischer2781:
You are right, that would probably be better. Closing this PR, this has been broken long enough that it doesn't need a hasty fix.

JaimeFrey:
I'm planning an overhaul and cleanup of the CE configuration. I don't known when it'll happen, but I'll make sure this is included in that effort.

PR #604, opened 2024-06-13
HTCONDOR-2501 Fix whole-node gpu routing to work with non-condor systems

JaimeFrey:
The existing expression for RequestGPUs will always evaluate to Undefined outside of HTCondor matchmaking against a slot ad.


matyasselmeci (approved):
LGTM

ColeBollig (approved):
LGTM... Not merging immediately due to possible other changes that may come about from more exploration of possible issues brought about in routes designed for targeting a backing HTCondor system but going into a slurm system

JaimeFrey:
I took one of the troubled job ads from the Hosted-CE we were investigating, and with this change RequestGPUs evaluated to the correct value. I don't know why we were seeing Undefined in some of the sub-expressions yesterday.

PR #607, opened 2024-07-18
HTCONDOR-2544: Update CE build and test to use OSG 23.

timtheisen:


timtheisen:
Can't update to Ubuntu 22.04 or 24.04. Doesn't work. Problem for another day.

matyasselmeci (approved):

PR #609, opened 2024-07-19
HTCONDOR-2549 daemon domain

JaimeFrey:
Use the local UID_DOMAIN (users.htcondor.org) for the identity of 'condor' clients. Otherwise, in 23.9.x, the schedd won't recognize the job router as a queue superuser.


ColeBollig (approved):
LGTM

PR #611, opened 2024-07-30
HTCONDOR-2563 batch-queue

JaimeFrey:
BatchQueue is the first-class way to specify the queue in the batch system. We should use it instead of queue or remote_queue.


ColeBollig (approved):
LGTM

PR #612, opened 2024-09-18
HTCONDOR-2637 Request old DN formatting for certificate issuer

JaimeFrey:
The openssl formats DNs differently in EL7, EL8, and EL9. Request the old format explicitly so that we know how to parse it.


timtheisen (approved):
LGTM

PR #618, opened 2024-10-25
HTCONDOR-2703 whole-node

JaimeFrey:
Setting expressions for the routed job's resource requests doesn't work properly if the routed job is going through the grid universe.


ColeBollig (approved):
LGTM

PR #619, opened 2024-11-12
HTCONDOR-2688 python-htcondor2

JaimeFrey:
condor_ce_trace required major changes due to altered or removed interfaces.

### DIFF
--- a/src/condor_ce_trace
+++ b/src/condor_ce_trace
@@ -11,8 +11,6 @@ import traceback
 
 os.environ.setdefault("CONDOR_CONFIG", "/etc/condor-ce/condor_config")
 
-#import classad
-#import htcondor
 import classad2 as classad
 import htcondor2 as htcondor
 import htcondorce.tools as ce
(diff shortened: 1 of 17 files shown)


src/condor_ce_trace
```diff
@@ -11,8 +11,10 @@ import traceback
 
 os.environ.setdefault("CONDOR_CONFIG", "/etc/condor-ce/condor_config")
 
-import classad
-import htcondor
+#import classad
+#import htcondor
```
ColeBollig, line 15:
```suggestion
```

ColeBollig (approved):
LGTM... there are test failures, but I believe @timtheisen is working on them. I guess don't merge until those are resolved (unless tim says otherwise)

ColeBollig:
I did some grepping and I believe that `src/audit_payloads.py` and `src/condor_ce_run` still need to be updated

JaimeFrey:
@ColeBollig What branch are you looking at. I don't even see an audit_payloads.py file.

timtheisen:
I believe that he was grepping around in master (which is ancient). Perhaps, main would be a better branch. @ColeBollig @JaimeFrey

ColeBollig:
@JaimeFrey, Tim is right... I was looking at `master` branch instead of `main`. My bad

timtheisen:
Will merge this and test with OSG VMU tests. Will circle back later to fix the Github actions

PR #621, opened 2024-11-22
HTCONDOR-2703 Use whole-node custom resource requests only for HTCondor

timtheisen:
Setting expressions for the routed job's resource requests doesn't work properly if the routed job is going through the grid universe.


ColeBollig (approved):
Looks the exact same

PR #622, opened 2024-11-22
HTCONDOR-2637 Request old DN formatting for certificate issuer

timtheisen:
The openssl formats DNs differently in EL7, EL8, and EL9. Request the old format explicitly so that we know how to parse it.


timtheisen:
Never mind. This fails on EL7. So, we'll live with it as is.

timtheisen:
Well, we only build for EL8 and EL9. So, this still applies

PR #624, opened 2024-12-18
Fix spots where we still want code block tabs

brianhlin:
Verified locally using:

```
podman run --rm -it -v ${PWD}:/docs -p 8000:8000 docker.io/squidfunk/mkdocs-material:7.1.0
```


timtheisen (approved):
LGTM

PR #634, opened 2025-05-22
Apel token accounting

vokac:
Basic APEL accounting for job submitted with tokens
Accounting for jobs without delegated VOMS proxy
Provide similar accounting options implemented in ARC-CE 7
(https://bugzilla.nordugrid.org/show_bug.cgi?id=4236).

* no changes for jobs with delegated VOMS proxy (x509UserProxyFirstFQAN)
* optionally use first wlcg.groups with APEL_USE_WLCG_GROUPS enabled
* mapfile for remaining job using Owner or AuthTokenIssuer,AuthTokenSubject

Functionality is limited to HTCondor-CE with local HTCondor.

### DIFF
--- a/contrib/apelscripts/condor_batch_blah.py
+++ b/contrib/apelscripts/condor_batch_blah.py
@@ -23,6 +23,7 @@ HTCondor configuration values:
     APEL_CE_ID           APEL identifier for the CE
     APEL_SCALING_ATTR    job attribute for optional performance factor
     APEL_SPEC_ATTR       job attribute for optional absolute performance
+    APEL_USE_WLCG_GROUPS extract accounting group from first wlcg.groups token claim
 """.strip(),
     formatter_class=argparse.RawDescriptionHelpFormatter,
 )
@@ -117,6 +118,10 @@ ce_host = read_ce_config_val("APEL_CE_HOST")
 ce_id = read_ce_config_val("APEL_CE_ID")
 scaling_expr = format_apel_scaling(options.apel_config, ce_id)
 output_datetime = time.strftime("%Y%m%d-%H%M")
+try:
+    use_wlcg_groups = read_ce_config_val("APEL_USE_WLCG_GROUPS").upper() == 'TRUE'
+except CalledProcessError:
+    use_wlcg_groups = False
 
 for directory in (history_dir / "quarantine", output_dir):
     directory.mkdir(exist_ok=True)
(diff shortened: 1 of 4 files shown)


vokac:
Detected flake8 errors are not related to this issue, somebody has to fix [htcondor:main](https://github.com/htcondor/htcondor-ce/tree/main) branch, most probably just remove useless `global` for reported errors

```
./src/htcondorce/info_query.py:41:5: F824 `global _logger` is unused: name is never assigned in scope
./src/htcondorce/web.py:95:5: F824 `global _plugins` is unused: name is never assigned in scope
```

timtheisen:
I have some work to get main to pass tests. It will probably have to wait until after HTC 25. I have a bunch of work to complete before then.

maarten-litmaath (commented):
In these lines, the userMap function is called twice with identical arguments, which may be expensive:

https://github.com/htcondor/htcondor-ce/pull/634/files#diff-11bab112429e5e1f254e9a6006fd6b8e169403c7cc84a506efcd7ce225a988bcR142-R143

contrib/apelscripts/condor_batch_blah.py
```diff
@@ -130,6 +135,13 @@ def condor_q_format(job_history: Path, *formats: str) -> str:
 else:
     batch_stream = blah_stream = sys.stdout
 
+accounting_user_expr = "ifThenElse(!isUndefined(x509userproxysubject), x509userproxysubject, orig_AuthTokenSubject)"
+accounting_group_expr = "ifThenElse(!isUndefined(x509UserProxyFirstFQAN), x509UserProxyFirstFQAN, %s)"
+if use_wlcg_groups:
+    accounting_group_expr = accounting_group_expr % 'ifThenElse(!isUndefined(orig_AuthTokenGroups), split(orig_AuthTokenGroups, ",")[0], %s)'
+accounting_group_expr = accounting_group_expr % 'ifThenElse(!isUndefined(userMap("ApelAGMap", Owner)), userMap("ApelAGMap", Owner), %s)'
```
maxfischer2781, line 142: If you want to express "if `A isnt undefined` then `A` else `B`" you can use the [ternary operator shorthand](https://htcondor.readthedocs.io/en/latest/classads/classad-mechanism.html#classad-operators) `A ?: B`. This is already used above in `format_apel_scaling` to build a similar cascade.
```suggestion
accounting_group_expr = accounting_group_expr % '(userMap("ApelAGMap", Owner) ?: %s)'
```

contrib/apelscripts/condor_batch_blah.py
```diff
@@ -130,6 +135,13 @@ def condor_q_format(job_history: Path, *formats: str) -> str:
...
+accounting_group_expr = accounting_group_expr % 'ifThenElse(!isUndefined(orig_AuthTokenIssuer) && !isUndefined(orig_AuthTokenSubject) && !isUndefined(userMap("ApelAGMap", strcat(orig_AuthTokenIssuer, ",", orig_AuthTokenSubject))), userMap("ApelAGMap", strcat(orig_AuthTokenIssuer, ",", orig_AuthTokenSubject)), %s)'
```
maxfischer2781, line 143: Similar to the above, you should be able to save the double `userMap` here too though it isn't quite as nice.
```suggestion
accounting_group_expr = accounting_group_expr % '(((orig_AuthTokenIssuer isnt UNDEFINED && orig_AuthTokenSubject isnt UNDEFINED) ? userMap("ApelAGMap", strcat(orig_AuthTokenIssuer, ",", orig_AuthTokenSubject)) : UNDEFINED) ?: %s)'
```

maarten-litmaath: I do not see why there need to be "(((" and "))" instead of "((" and ")" ?
Also, is the ": UNDEFINED" really needed?

maxfischer2781: I'm not entirely sure about the precedence rule of ClassAd operators, so the extra parentheses are to be conservative. I guess the pair around the first condition could be removed.

The `: UNDEFINED` is needed to have the full ternary – something must be produced in that branch, and the next ladder step expects a result or `UNDEFINED`. Sadly, [`strcat` evaluates to `ERROR` on `UNDEFINED` input](https://htcondor.readthedocs.io/en/latest/classads/classad-builtin-functions.html#strcat()) so one can't just use the `?:` operator as before.

FWIW, since the next step is a constant `UNDEFINED` anyways, one could skip the logic for inserting a placeholder branch and just stop at `UNDEFINED`. It would also probably look a bit nicer if the expression were built in reverse (as I did in `format_apel_scaling`) which allows to automatically create the scaffolding for the ladder, without hardcoding it into every step.

JaimeFrey (approved):
Looks good to me.

